PR #17, opened 2019-10-31
WIP: Vendor chunk

rmader:
Let Webpack help us optimizing our JS. Reduces combined size of vendor.js + adhocracy.js quite a bit, and some other endpoints, especially map related ones, aswell.

Depends on https://github.com/liqd/adhocracy4/pull/360


rmader:
More todos:
 - drop `timeago.js` in favour of `Intl.RelativeTimeFormat` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RelativeTimeFormat
 - update webpack to only serve to modern browsers - we don't need to support IE11, which is the only still supported legacy browser - all other legacy browsers are unsafe and should not be used *AT ALL*

PR #18, opened 2019-11-04
assets/embed: add outside border to embed and change frame colour, ad…

phillimorland:
…d non overwritable colour variable for a+ logo colour


adhocracy-plus/assets/scss/_variables.scss
```diff
@@ -5,6 +5,7 @@ $secondary: #00a747 !default;
 $brand-primary: #008dc5 !default;
 $brand-secondary: #2d40cc !default; //same as $brand-info
 $brand-danger: #a72b1e !default;
+$platform-logo: #90efe4 !default; //same as secondary, to be used in org when we still need the turquiose
```
fuzzylogic2000: I've got the feeling, that we should scrap the organisation colours. They are much work, and not much loved, it seems to me. But that's for another day to discuss.

PR #77, opened 2019-11-12
Mn 2019 11 change login and registration forms

### LINKED ISSUE #36: Login: Change label with placeholder text
The label for the email address / username field should be the placeholder text, which is more explicit and clear

https://product-dev.liqd.net/accounts/login/?next=/

bildschirmfoto 2017-11-28 um 11 32 13

### LINKED ISSUE #75: registration form: add help text to username field
We should make users who register aware of the fact that their username will be publicly shown by adding a little help text underneath the username form label. We can use the same text as in https://mein.berlin.de/accounts/signup/?next=/

MagdaN:
fixes #75 
fixes #36

### DIFF
--- a/apps/users/forms.py
+++ b/apps/users/forms.py
@@ -1,3 +1,4 @@
+from allauth.account.forms import LoginForm
 from allauth.account.forms import SignupForm
 from allauth.socialaccount.forms import SignupForm as SocialSignupForm
 from django import forms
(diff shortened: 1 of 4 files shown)


apps/users/forms.py
```diff
@@ -26,8 +36,23 @@ class SocialTermsSignupForm(SocialSignupForm):
...
         user.get_newsletters = self.cleaned_data['get_newsletters']
         user.save()
         return user
+
+
+class ExtraLabelLoginForm(LoginForm):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields['login'].label = _('username/e-mail address')
```
fuzzylogic2000: I think this is meant to be 'Username or e-mail'?!

PR #104, opened 2019-11-20
apps/projects: fix next in participant project invite

fuzzylogic2000:
I think, this should fix the NoReverseMatch. But I couldn't reproduce it. Will try again, but maybe you could as well @rmader or @MagdaN ?


fuzzylogic2000:
Alright, I can't reproduce the sentry error, but prior to these changes the participant invite was broken for sure. @rmader Should we merge and deploy anyway?

PR #115, opened 2019-11-26
hompage hero: add multiple image chooser to homepage and style

phillimorland:
Not sure why MK's commit is in with mine? 
also i did a js thing, hope it's ok


MagdaN:
OK I removed the md linting

PR #150, opened 2019-12-10
Multi-site improvements

rmader:
This extends multi-site support for emails quite a bit.


rmader:
Test fixed by setting the `SITE_ID` only for test. @fuzzylogic2000 @MagdaN: are you ok to let land this on master so we can test on dev?

fuzzylogic2000:
Yes, let's do that!

PR #162, opened 2019-12-11
Pm 2019 12 rm comments

phillimorland:
not displaying comment box due to undefined default rating


adhocracy-plus/assets/js/app.js
```diff
@@ -47,7 +48,8 @@ var initialiseWidget = function (namespace, name, fn) {
 }
 
 var init = function () {
-  initialiseWidget('a4', 'comment', ReactComments.renderComment)
+  initialiseWidget('a4', 'comment_async', ReactComments.renderComment)
+  initialiseWidget('a4', 'comment_with_categories', ReactCommentsWithCategories.renderComment)
```
rmader: Just a nit, but can we make it `comment_async_with_categories` then? To make it absolutely obvious that they belong together.

PR #165, opened 2019-12-12
sitemaps: Make sitemaps organisation specific and add multi-site support

rmader:
This moves organisation and project specific sitemaps to the organisation
app, allowing it to be reachable in multi-site setups. Further more, it
makes it easier to add things like the imprint and other static sites.

---

~~This is *almost* ready to go, but the is one issue left where a sitemap is liked as url, which is not completely trivial to solve :(~~


rmader:
Should be ready now

PR #205, opened 2020-01-06
WIP assets/as-comments: comment nav bar sit on single row

phillimorland:


rmader:
Did the A4 part for this land already? Would like to deploy A+ today, should we land this for that?

rmader:
Ah never mind, will release now, we can add this in the next one.

PR #229, opened 2020-01-23
Mn 2020 project sorting on org pages

MagdaN:


apps/organisations/models.py
```diff
@@ -94,6 +97,45 @@ class Organisation(models.Model):
...
+            .filter(organisation=self,
+                    is_archived=False,
+                    is_draft=False)
+
+    def get_projects_list(self, user):
+        active_projects = []
+        future_projects = []
+        past_projects = []
+
+        projects = query.filter_viewable(self.projects, user)
```
fuzzylogic2000: What is the difference to self.projects? Is it sth. from the rules, so that also private projects are displayed?

apps/organisations/templates/organisation_landing_page.html
```diff
@@ -48,30 +48,44 @@ <h3 class="u-no-margin organisation__slogan">{{ organisation.slogan }}</h3>
...
-                        </li>
-                    {% endfor %}
-                </ul>
+            {% if object.projects.count == 1 %}
+            <div class="col-lg-12">
+              {% has_perm 'a4_candy_projects.view_project' request.user object.projects.0 as can_view_project %}
+              {% if can_view_project  %}
+              {% include 'a4_candy_projects/includes/project_list_tile.html' with project=object.projects.0 orientation='horizontal' %}
             </div>
+            {% endif %}
```
fuzzylogic2000: endif
else
I think, it's the indentation that's strange?!

apps/organisations/views.py
```diff
@@ -19,29 +16,12 @@ class OrganisationView(DetailView):
...
-            .exclude_updates()[:4]
-
-        context['stats'] = {
-            'users': 1204,
-            'items': 3425,
-            'comments': 23234,
-            'ratings': 134234,
-        }
-
+        context['active_projects'] = active
```
fuzzylogic2000: Nice to put it all into the model!

PR #231, opened 2020-01-23
Ka 2020 01 restyle debate question tiles

fuzzylogic2000:


phillimorland:
I commented out the demi bold for now as it displaying incorrectly, will investigate now also increased header size to be serif to align with design but will also speak to Marcello about that

PR #235, opened 2020-01-24
apps/debate: fix comment export and add subject export with comment c…

fuzzylogic2000:
…ount


fuzzylogic2000:
@phillimorland Now it's fine! Or should be.

PR #240, opened 2020-01-27
assets: Big app.js / webpack cleanup from mB

phillimorland:
And some other fixes around embed. Fixes a couple of bugs, like
the polygon download on the way. Sorry for the single commit.

@rmader this broke stuff on dev but not locally so we reverted and hopefully we can fix on this branch


rmader:
Closed in favour of #247

PR #256, opened 2020-01-30
Pm 2020 01 org cols and social media

phillimorland:
still awaiting final wording on social media heading


adhocracy-plus/assets/scss/_layout.scss
```diff
@@ -52,3 +56,10 @@
     z-index: 1;
     top: -1em;
 }
+
+.l-two-col {
+    -moz-column-count: 2;
+    -moz-column-gap: 2*$r-spacer;
+    -webkit-column-count: 2;
+    -webkit-column-gap: 2*$r-spacer;
+}
```
phillimorland: @fuzzylogic2000 my brain is crap today but i'm hoping we can just use this but its not covered by all browsers https://caniuse.com/#feat=multicolumn what do you think?

fuzzylogic2000: Yeah, that doesn't sound too bad and is working with the popular browsers, so let's do it!

fuzzylogic2000:
@phillimorland Alright! Could you have a look and merge if fine?

PR #300, opened 2020-02-10
config/base: rm english language including migration

phillimorland:


fuzzylogic2000:
Let's try the hacky way: #308 :(

PR #308, opened 2020-02-11
languages: force site to be in german, hide switch from header, and h…

fuzzylogic2000:
…ide form field from user settings

Uh, this is quite a hack, but seems to be the easiest solution with no migrations or changes to the tests.
It forces all templates to be German with the force-templatetag, 
hides the switch from the header,
and hides the form field from the user settings. I did hide that and didn't remove it from the form, because the latter would also mean that we needed to change the form_valid in the views.


fuzzylogic2000:
@rmader I fixed the linting error and rebased. What do you say?

PR #318, opened 2020-02-12
apps/organisations: add members

fuzzylogic2000:
Adds a member model to the organisations app.

Next steps:
- make the inline-members-thing for organisations in the django-admin use the one-to-many-widget (I am currently very stuck on that, and will give up for now. Maybe we don't need that anyway)


apps/organisations/admin.py
```diff
@@ -3,7 +3,27 @@
 from . import models
 
 
+class MemberInline(admin.TabularInline):
+    model = models.Member
+    fields = ['member',]
```
rmader: polylint says:
`apps/organisations/admin.py:8:23: E231 missing whitespace after ','`

fuzzylogic2000: right! That was the thing I also wanted to investigate. :roll_eyes:

PR #321, opened 2020-02-13
Mn 2020 02 add captcha

MagdaN:


adhocracy-plus/assets/js/captcheck.js
```diff
@@ -0,0 +1,121 @@
+window.onload = function () {
```
rmader: Is this file mostly copied? Then we should add the license text

MagdaN: Yes, I just changed it in a way that the api url is not hard coded and adapted it to our linting rules. Will add the license

MagdaN:
@rmader then I will remove the app from the setting and the registration form (I just added it for testing)

rmader:
Would it be possible to check in the registration form if its activated? Just like we check for the servicekonto app?

PR #326, opened 2020-02-13
Add custom registration form for IG-BCE

rmader:
Not quite done yet

### DIFF
--- a/adhocracy-plus/config/settings/base.py
+++ b/adhocracy-plus/config/settings/base.py
@@ -249,8 +249,8 @@ ACCOUNT_AUTHENTICATION_METHOD = 'username_email'
 ACCOUNT_EMAIL_CONFIRMATION_EXPIRE_DAYS = 3
 ACCOUNT_EMAIL_REQUIRED = True
 ACCOUNT_EMAIL_VERIFICATION = 'mandatory'
-ACCOUNT_FORMS = {'signup': 'apps.users.forms.TermsSignupForm',
-                 'login': 'apps.users.forms.ExtraLabelLoginForm'}
+ACCOUNT_FORMS = {'signup': 'apps.users.forms.DefaultSignupForm',
+                 'login': 'apps.users.forms.DefaultLoginForm'}
 ACCOUNT_LOGIN_ATTEMPTS_LIMIT = 10
 ACCOUNT_LOGIN_ATTEMPTS_TIMEOUT = 300  # seconds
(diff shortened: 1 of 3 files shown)


apps/custom_auth/forms.py
```diff
@@ -0,0 +1,132 @@
...
+            raise forms.ValidationError(
+                "Something is wrong with the setup - please try again later"
+            )
+
+        try:
+            Member.objects.get(member_number=member_number)
+            raise forms.ValidationError(
+                "A member with that number already exists."
+            )
+        except ObjectDoesNotExist:
```
fuzzylogic2000: I know, I don't strictly need the except, but having just the try makes me cringe.

apps/custom_auth/forms.py
```diff
@@ -0,0 +1,132 @@
...
+        birth_date = cleaned_data.get("birth_date")
+
+        self.validateMemberNumberAndDate(member_number, birth_date)
+
+    def save(self, request):
+        user = super(IgbceSignupForm, self).save(request)
+        organisation = Organisation.objects.get(
+            id=settings.SITE_ORGANISATION_ID
+        )
+        additional_info = {
```
fuzzylogic2000: do I need to tell it, that it has to be json or is putting a dict into a json-field enough?

apps/custom_auth/forms.py
```diff
@@ -0,0 +1,132 @@
...
+            )
+
+    def clean(self):
+        cleaned_data = super().clean()
+        member_number = cleaned_data.get("member_number")
+        birth_date = cleaned_data.get("birth_date")
+
+        self.validateMemberNumberAndDate(member_number, birth_date)
+
+    def save(self, request):
```
fuzzylogic2000: do we need to make sure that all the saving here is atomic or does the save already handle that?

apps/custom_auth/forms.py
rmader: `Member.objects.filter(member_number=member_number).exists()` should be cleaner

apps/custom_auth/forms.py
rmader: Don't see a race condition here...we can drop the `user.save()` though

apps/custom_auth/forms.py
rmader: And, to add more smartassing: `Member.DoesNotExist` ;)

fuzzylogic2000: Ah! That's the way!

PR #439, opened 2020-03-02
cms/page: add migration to create homepage

phillimorland:
When there is a new wagtail page, the "Welcome to your ...." has always the page id 2.
So, when there isn't anything added to wagtail yet, it should still be there and have no child pages. In that case, we run the stuff in this migration (and delete the wagtail page and add a new page and set that as home).
But there can be three different scenarios, where or how the wagtail pages where changed:
- A new homepage was added, the old wagtail page was deleted (like that on all aplus instances). This should be kept like that, but will be, because `Page.objects.filter(id=2).exists()` is False.
- A new child page was added to the existing wagtail-homepage (like that on the enquete-instances). This should also be kept like that ((but shouldn't ever happen again)), and will be because `Page.objects.get(id=2).numchild == 0` is False.
- A new homepage was added, but the old wagtail page was kept. Then the new homepage has to be added as a child to the wagtail homepage and the site has to be set differently, so this is a very weird setup that shouldn't be like that anywhere (but was the case on my local instance of a+ and could be like that elsewhere). So, this should also be kept like that and will be bacause `Page.objects.get(id=2).numchild == 0` is False.

I hope, I could think of all possibilities how this could break anything and worked around that. Could you please test it on your local DBs? @phillimorland @Rineee @rmader


phillimorland:
@fuzzylogic2000 I can't review but it's working with my database! :)

fuzzylogic2000:
@rmader Could you also try this and merge if works fine?

rmader:
Nice @phillimorland ! I'm slightly worried that we do not have a direct dependency on the migration that adds the wagtail page, but it looks like it is one of the `wagtailcore` ones and in our initial migration we depend on `wagtailcore 0041_group_collection_permissions_verbose_name_plural` - so we *should* have a indirect one. So we should be save here. Great, looks good to me!

PR #478, opened 2020-03-09
deps: Remove deprecated popper.js

rmader:
Apparently a newer version is either installed as dependecy or it
is bundled in - either way, this package version got deprecated and
won't install any more and things appear to work just fine without it.


rmader:
@phillimorland: do you know what this was needed for?

PR #491, opened 2020-03-17
Pm 2020 03 comments test

phillimorland:
this is branch to be tested alongside https://github.com/liqd/adhocracy4/pull/504


apps/contrib/templates/a4_candy_contrib/item_detail.html
```diff
@@ -140,7 +140,7 @@ <h2 class="moderator-statement__title">{% trans 'Official Statement' %}</h2>
 
     <div class="row">
         <div class="col-md-8 offset-md-2">
-            {% react_comments object %}
+            {% react_comments_async object False %}
```
fuzzylogic2000: you don't need the false, I think, because you already put that into the tag.

phillimorland: ok cool! i'll test that and squash and push again, thank you!!!!

phillimorland: I tried removing the false and it just showed the categories everywhere so I left it in for now while marcelo has a look and then we can look at it a bit more closely when we do styling?

PR #494, opened 2020-03-23
WIP: cms/blocks.py: remove choice option to choose cols and update templat…

phillimorland:
…e incl migration
@fuzzylogic2000 when I got into this, i found that a lot of what I did was very messy and would like to refactor a lot but for right now, I just updated that block. I'm not sure however if this is the best way to go about doing it? it is kind of done this way in ae/ce but I would like your opinion


fuzzylogic2000:
Hmm, it's reusabilty vs. making blocks exactly for what they are meant to do, right? I would go a bit more in the direction of the latter. Maybe reusabilty makes sense for some columns, but if the styles are very restrictive and the blocks have to be set up a certain way for that to work, I think it's easiest if we just allow the certain way for the whole block.

phillimorland:
@fuzzylogic2000 i messed up migrations by renaming, i'll redo and tag you

phillimorland:
@fuzzylogic2000 what do we need to do on this now?

fuzzylogic2000:
@phillimorland I still need to get the migrations right, but don't know how to go on about it. I think, I need to test what we need and how to get the old stuff into the new models.

phillimorland:
ok if you want to pair on it hit me up!

fuzzylogic2000:
will hopefully replaced by #1295

PR #513, opened 2020-04-02
apps/organisations: update edit setting template to extend dashboard,…

phillimorland:
… rm from urls.py and mv to dashboard, update base_project_list to mv settings btn to dashboard tab

@fuzzylogic2000 there you go :)


fuzzylogic2000:
That fixed the tabs, now there's the test left to fix. Will do!

fuzzylogic2000:
So, and these are the tests. I think, when you looked at it @phillimorland we should squash all three commits. So just give me a shout if you did!

phillimorland:
@fuzzylogic2000 coolio! danke! i squash and add the project list stuff

adhocracy-plus/templates/a4dashboard/base_project_list.html
```diff
@@ -21,10 +21,6 @@
 
         {% has_perm 'a4_candy_organisations.change_organisation' request.user view.organisation as user_may_change_organisation %}
         {% if user_may_change_organisation %}
-            <a href="{% url 'organisation-settings' organisation_slug=view.organisation.slug %}"
-               class="btn btn--light btn--full">
-                {% trans 'Edit your organisation' %}
-            </a>
             {% if settings.a4_candy_cms_settings.ImportantPages.manual_link %}
```
fuzzylogic2000, line 24: Nothing to do with your PR, but we should show the link to everyone that can access the dashboard, not only to the ones that are allowed to change the organisation.

fuzzylogic2000: Ah, let's do that later! We'll look at the dashboard for the whole sprint now!

phillimorland: ok i'm doing some last cleaning. why should people that can't click on the link be able to see it though? or am i mis-understanding the permissions?

fuzzylogic2000: The link only links to the manual. Our a+-wiki page. Anyone should be able to see it.
I think the permission was there for the organisation update link (where it makes sense), but wasn't closed in the right place. Or is there another link even after that?

PR #515, opened 2020-04-02
templates/header//footer: adding additional lower header and upper fo…

phillimorland:
organisation header i'm still having a play with as we have no mobile design yet, but so far it's ok on the main pages, sub pages need a bit more work but would be great if can merge this so Marcelo can have a proper look and the multimodule can be looked into


phillimorland:
@fuzzylogic2000 yes please! I should have tagged you in description, it says that there, sorry!

fuzzylogic2000:
Haha, yes, reading the dexription would have neem a good idea...

PR #517, opened 2020-04-05
Ks 2020 04 change organisation model

Rineee:


apps/organisations/models.py
```diff
@@ -112,7 +114,23 @@ class Organisation(models.Model):
...
+    )
+    data_protection = RichTextField(
+        verbose_name=_('Data protection'),
+        help_text=_('Please provide your data protection information.'),
+        blank=True
+    )
+    netiquette = RichTextField(
+        verbose_name=_('Netiquette'),
+        help_text=_('Please provide your rules for online discussions.'),
+        blank=True
```
Rineee, line 133: I wasnt sure about the help text, should I ask Frede about them?

adhocracy-plus/templates/header.html
```diff
@@ -40,6 +40,10 @@
                             {% url 'organisation-imprint' organisation_slug=ORGANISATION.slug as organisation_imprint_url %}
                             <a href="{{ organisation_imprint_url }}" class="organisation-header__nav-link"><span class="{% if request.path == organisation_imprint_url %}is-active{% endif %}">{% trans 'Imprint' %}</span></a>
                         </li>
+                        <li class="organisation-header__nav-item">
+                            {% url 'organisation-terms-of-use' organisation_slug=ORGANISATION.slug as organisation_terms_of_use_url %}
+                            <a href="{{ organisation_terms_of_use_url }}" class="organisation-header__nav-link"><span class="{% if request.path == organisation_terms_of_use_url %}is-active{% endif %}">{% trans 'Terms of use' %}</span></a>
+                        </li>
```
Rineee, line 46: This is actually just to test the new template until clear where it should be linked..

apps/organisations/models.py
fuzzylogic2000: About the name and helptext? Yes, please ask Frede! For the German and English text. Also I think in Englilsh it's something else and you also need to update the field name. :)
And if you have done so, you can just delete your old new migration and make a new one, so that all changes are in one migration. Just make sure you have a fresh DB, before making the new migrations.

adhocracy-plus/templates/header.html
fuzzylogic2000: Cool! I think @phillimorland already started with updating the footer, so you can talk to her about where it needs to go (or leave out the links and just put the urls in the description of the PR) and then put the links somewhere after merging.

Rineee:
Urls for the templates are also added, so they are ready to be linked in the footer..

apps/organisations/models.py
```diff
@@ -28,13 +28,15 @@ class Organisation(models.Model):
         max_length=100,
         default='Organisation',
         help_text=_('The title of your organisation will be shown '
-                    'on the landing page. max. 100 characters')
+                    'on the landing page. max. 100 characters'),
+        blank=True
```
phillimorland, line 32: I think the organisation name should be required

Rineee: But this is the title, the name is still required..I agreed on that with Katharina, see https://taiga.liqd.net/project/liqd-liquid-software/us/2796?milestone=214

phillimorland: ok cool, maybe we can add required to the html or something

PR #518, opened 2020-04-06
Ks 2020 04 split up organisation form

Rineee:


apps/organisations/templates/a4_candy_organisations/organisation_form.html
```diff
@@ -1,12 +1,12 @@
...
 {% load i18n %}
 
 {% block title %}{% trans "Settings" %} &mdash; {{ block.super }}{% endblock %}
 
-{% block dashboard_content %}
+{% block organisation_form %}
 <div class="container">
     <div class="row">
-        <div class="col-md-10 col-lg-8 offset-md-1 offset-lg-2">
+        <div class="col-md-10 col-lg-10 offset-md-1">
```
Rineee, line 9: @phillimorland I am not so sure what I did here :) Before the form was really narrow and it looked better like this, but maybe you can check again if this makes sense?

phillimorland: @Rineee nice! the class names dictate the amount of columns the content of the div will take up at each size of screen so now this will be 10 columns wide on both medium and large and the offset centers the content. As bootstrap uses the last class name for all bigger screens we don't need the extra col-lg-10 so i'll delete it :)

Rineee: Ok cool, thank you for the explanation!!

apps/organisations/forms.py
```diff
@@ -7,12 +7,22 @@ class OrganisationForm(forms.ModelForm):
...
+
+class OrganisationLegalInformationForm(forms.ModelForm):
+
+    class Meta:
+        model = Organisation
+        fields = ['imprint', 'terms_of_use', 'data_protection',
+                  'netiquette']
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
fuzzylogic2000, line 28: You shouldn't need this here. Only when you overwrite something, which you don't here,

PR #532, opened 2020-04-09
Ks 2020 04 add organisation footer to emails

Rineee:


adhocracy-plus/config/settings/dev.py
```diff
@@ -31,3 +31,4 @@
 
 BASE_URL = 'http://localhost:8004'
 CAPTCHA_URL = 'https://captcheck.netsyms.com/api.php'
+SITE_ID = 1
```
fuzzylogic2000, line 34: good idea to have that in the dev-settings!

PR #545, opened 2020-04-15
Pm 2020 04 mobile menu

phillimorland:


adhocracy-plus/assets/scss/components/_header_lower.scss
```diff
@@ -54,19 +54,28 @@
...
 }
 
+// .header-lower__nav-link {
+//     display: inline;
+//
+//     &:hover {
+//         color: $brand-primary;
+//     }
+// }
+
```
phillimorland, line 68: i still might need this so left it in till i do my clean up pull request

PR #550, opened 2020-04-16
Ka 2020 04 legal links

fuzzylogic2000:


adhocracy-plus/templates/footer_upper.html
```diff
@@ -6,12 +6,27 @@
...
         <ul class="navbar-nav">
             <li class="nav-item">
-              {% url 'organisation-imprint' organisation_slug=ORGANISATION.slug as organisation_imprint_url %}
-              <a href="{{ organisation_imprint_url }}" class="header-lower__nav-link"><span class="{% if request.path == organisation_imprint_url %}is-active{% endif %}">{% trans 'Imprint' %}</span></a>                    </li>
+                {% url 'organisation-imprint' organisation_slug=ORGANISATION.slug as organisation_imprint_url %}
+                <a href="{{ organisation_imprint_url }}" class="nav-link header-lower__nav-link"><span class="{% if request.path == organisation_imprint_url %}is-active{% endif %}">{% trans 'Imprint' %}</span></a>
+            </li>
+            <li class="nav-item">
+                {% url 'organisation-data-protection' organisation_slug=ORGANISATION.slug as organisation_data_protection_url %}
+                <a href="{{ organisation_data_protection_url }}" class="nav-link header-lower__nav-link"><span class="{% if request.path == organisation_data_protection_url %}is-active{% endif %}">{% trans 'Data protection policy' %}</span></a>
```
fuzzylogic2000, line 18: This adds "is-active" when you are on the site, but the bootstrap way would be to add "active" to the li, but that also doesn't change the styling. So I would remove the active-class for now?

phillimorland: yeah I think thats fine!

PR #578, opened 2020-04-28
update dependencies: update a4 hash to fix comment edit

phillimorland:


fuzzylogic2000:
Was this the first try? I am confused about the hashes?!

fuzzylogic2000:
Yes, was!

PR #588, opened 2020-05-04
Pm 2020 04 dashboard left

phillimorland:
@fuzzylogic2000 this currently works as there is only 1 module but when there are more then 1 I'll need to add unique id to module menus


fuzzylogic2000:
> @fuzzylogic2000 this currently works as there is only 1 module but when there are more then 1 I'll need to add unique id to module menus

I already have multi moduole projects and there will be others, as the button was already enabled for initiators in a+... But I will still merge, because for now it doesn't matter if all module dropdowns are openend and closed on the same click!

PR #598, opened 2020-05-06
templates/language_switch: replacing lang switch with dropdown and ke…

phillimorland:
…eping it a11y compliant
It's working but this is first step in header refactor as it's a bit of a mess so this might be replaced or followed by full refactor


phillimorland:
closed in favor of https://github.com/liqd/adhocracy-plus/pull/600

PR #599, opened 2020-05-06
templates/base_dashboard: add functionality to toggle and dropdown bt…

phillimorland:
…ngroup


adhocracy-plus/assets/scss/components/_button.scss
```diff
@@ -229,7 +229,9 @@
     border-radius: 6.25rem;
 }
 
-.btn-group {
+.btn--group {
+    @extend .btn--group;
```
fuzzylogic2000, line 238: Should one of these be with one dash? Or why is it extending itself?

PR #670, opened 2020-06-25
WIP: email translations

rmader:
This is an attempt to move email translation into the generic locale files. I'm not completely happy with it yet, but it work fine. So this is in a request for comments stage.


Rineee:
Yay, supercool and seems pretty straightforward.. what are you not happy about?

rmader:
So with this solution we still need "dummy" templates for each supported language. But that is arguably a minor annoyance.

So I guess we should discuss together if this a desirable solution and if we agree that it is - well, then the research story can be closed :)

Rineee:
I think I found a way to avoid the multiple templates: we can override the EmailBase.render() from A4 in AplusEmail. There, we can then always use the english template, but choose the language for translation. I already did that and it works. Then, the question is, which language to use? If defined: user language of receiver, else language that is set by language switch, else default language (is currently german?)?
Does that sound good to you? I would then go in this branch..

Rineee:
fixes #673

Notifications/notify_followers_event_upcoming.en.email is not tested yet.. there I did another attempt to combine blocktrans and htlm links (in block reason). If that works, maybe the link generation in users/emails.py can be removed again?

Furhtermore, in most other emails I found (some variaton of) the sentence 'if you dont want to receive.. change your account settings', but without a link. I left it like this for now, but maybe that should also be links for consistency? Which would make NotificationsEmailBase redundant again :)

phillimorland:
@Rineee @rmader is this the first branch which then became the completed email translations? and as such can we close? same question for these https://github.com/liqd/adhocracy-plus/pull/704 https://github.com/liqd/adhocracy-plus/pull/705

rmader:
Yes, this can be closed

PR #685, opened 2020-07-08
README: Add deployment documentation

rmader:
Todo:
- install guide
- update guide
  - extra `make` commands?
- maybe some cleaups


rmader:
I think this is okayish for now - lets put it in

PR #704, opened 2020-07-21
Ks 2020 07 merge translations

Rineee:
So this is it, email translations with merged translation files!


Rineee:
And it cannot be rebased due to conflicts.. ah!

phillimorland:
@Rineee this too?

Rineee:
yes!

PR #731, opened 2020-08-10
WIP DIID Pm 2020 07 footer

phillimorland:
not including translations - need to decide how we want to do translations?


fuzzylogic2000:
I think logo and colours was also done in a different PR?!

PR #737, opened 2020-08-11
requirements: update isort to 5.2.3, travis.yml: update travis config to be inline with other projects

phillimorland:
Ok what I think is happening is:
Lint script has been moved to package.json so it can run make lint.
A script to run the lint-staged config has been added and this is called in make lint-quick. Pre-commit calls make lint-quick which runs the lint-staged config on required files

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,8 @@
   "scripts": {
     "build:prod": "webpack --config webpack.prod.js --mode production",
     "build": "webpack --config webpack.dev.js --mode development",
-    "watch": "webpack --config webpack.dev.js --watch --mode development"
+    "watch": "webpack --config webpack.dev.js --watch --mode development",
+    "lint": "eslint apps adhocracy-plus/assets --ext .js,.jsx && stylelint 'adhocracy-plus/assets/scss/**/*.scss' --syntax scss"
   },
   "browserslist": "defaults and not dead and >= 0.5%, ie >= 11",
   "husky": {
--- a/package.json
+++ b/package.json
@@ -67,8 +67,8 @@
     "eslint-plugin-react": "7.20.5",
     "eslint-plugin-standard": "4.0.1",
     "husky": "4.2.5",
+    "lint-staged": "10.2.11",
     "markdownlint-cli": "0.23.2",
-    "polylint.sh": "0.0.4",
     "stylelint": "13.6.1",
     "stylelint-config-standard": "20.0.0",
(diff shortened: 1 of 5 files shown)


rmader:
The polylint package can now get removed from package.json

rmader:
Also, the `lint` and `lint-quick` commands in `Makefile` will need some adoption. The `lint-quick` command is also set up as pre-commit hock in the `husky` section in `package.json`, in case we just want to delete in (like in meinBerlin).

However, running a full lint on every commit is quite slow - so it would be super cool if it was only run on changed files - but I have no idea how that was archived in `polylint` -> optional :)

phillimorland:
just quickly looking it up I found lint-staged https://github.com/okonet/lint-staged which could work, i'll try setting it up

phillimorland:
@rmader I ended up adding it to scripts otherwise it wasn't running but I think it's now working?!

package.json
```diff
@@ -67,8 +67,8 @@
     "eslint-plugin-react": "7.20.5",
     "eslint-plugin-standard": "4.0.1",
     "husky": "4.2.5",
+    "lint-staged": "10.2.11",
```
rmader, line 70: IIUC, we don't use that now, right? Because you just build our own `lint-staged` command.

package.json
```diff
@@ -77,7 +77,9 @@
   "scripts": {
     "build:prod": "webpack --config webpack.prod.js --mode production",
     "build": "webpack --config webpack.dev.js --mode development",
-    "watch": "webpack --config webpack.dev.js --watch --mode development"
+    "watch": "webpack --config webpack.dev.js --watch --mode development",
+    "lint": "eslint apps adhocracy-plus/assets --ext .js,.jsx && stylelint 'adhocracy-plus/assets/scss/**/*.scss' --syntax scss",
+    "lint-staged" : "eslint apps adhocracy-plus/assets --ext .js,.jsx --fix-dry-run"
```
rmader, line 82: This looks to me like it would only defer from `lint` in that in adds `--fix-dry-run` and doesn't run `stylelint`, but I don't see it only linting staged files :/

package.json
phillimorland: err no I don't, I just tell it where to look i thought, but don't worry, i'll read the docs properly and i'll come back to this after I've got some issues fixed

rmader:
Closed in favour of https://github.com/liqd/adhocracy-plus/pull/757

PR #757, opened 2020-08-18
Pm 2020 isort test

phillimorland:
I am not convinced it's faster but it does seem to be doing what it's meant to, my script might be wrong though?!
Oh and need to do a make install, i forgot that when coming back to test :p


rmader:
@phillimorland: Great! Took the freedom to add some fixes for the python part which didn't quite work.

fuzzylogic2000 (approved):
Yeah!

phillimorland:
@rmader at some point can you explain the things you added so I know for next time? :)

PR #760, opened 2020-08-19
diid: remove adhocracy+ from footers

fuzzylogic2000:


rmader (approved):
Nice. Would still be great to get all content out of the templates somehow in the future, but thats more of an epic

PR #769, opened 2020-08-25
docs: Some additions to shibboleth config

rmader:
- xmlsec path must be set for djangosaml2 to work
- nodes disables key password protection, which is not helpful in
this scenario


rmader:
This shouldn't need a review

PR #773, opened 2020-08-27
apps/questions: Use organisation slug

rmader:
So we don't need nginx rewrites to make this work for organisations
with own domains.


rmader:
FTR: I already applied that to enquete-master and enquete-release, see e.g. https://stage.enquetebeteiligung.de/projects/module/interaktive-veranstaltung/

fuzzylogic2000 (approved):
Yeah!

PR #808, opened 2020-09-22
apps/questions: style module info and phase info to be more readable

phillimorland:
small changes for Monday if possible


fuzzylogic2000 (approved):
Nice!

PR #815, opened 2020-10-05
Ks 2020 09 multi language organisation form

Rineee:
To do: adjust tests


fuzzylogic2000 (approved):
Yes! It reads well, and like you thought of everything! And you even found the css-variable! And it also works!

PR #823, opened 2020-10-06
Matomo tracking

rmader:
Enable the ability to add matomo tracking and add no-cookie option.


fuzzylogic2000 (changes requested):
Looks good to me - and was tested in mB, right? But travis doesn't like it. Looks like there is setting missing in test settings?

rmader:
Tests are fixed now, also added some more cleanups :)

PR #826, opened 2020-10-07
apps/organisations: remove old untranslated fields

Rineee:
Do not merge this until the new translatable fields are deployed!


fuzzylogic2000 (approved):
Looks good! We can merge after release (and checking that all fields are there in the German fields)!

fuzzylogic2000:
It all worked well locally, and the fields are migrated fine on dev, stage and prod. So, let's go!

PR #831, opened 2020-10-08
WIP: tests/organisations: test update translated fields

fuzzylogic2000:


fuzzylogic2000:
Maybe there's a better way to test the locale the emails are sent in? But I didn't find it.
@rmader @Rineee Any ideas?

rmader:
Looks good to me! Do we need the binary files though?

fuzzylogic2000:
@rmader Yes, we do, because they are the actual translations and no need to run compilemessages then. (I found it like that in the django tests and it worked!)

Rineee (approved):
I think its really cool!

PR #832, opened 2020-10-12
settings: disable parler caching

### LINKED ISSUE #827: # 2929 testing deleting languages (language availability)
**URL:** https://aplus-dev.liqd.net/teststadt-luca/dashboard/settings/
**user:** initiator
**expected behaviour:** deleting languages in the organization's settings
**behaviour:** sometimes the language buttom does not disappear when trying to delete a language. Other times an error 500 occurs when trying to delete a language. 
**important screensize:** all
**device & browser:** all
**Comment/Question:** I could not identify a pattern when that happens, but it appears to be related to the selection of the default language for E-mails.

Rineee:
lets hope, that this was the problem! it works on local at least..

fixes #827

### DIFF
--- a/adhocracy-plus/config/settings/base.py
+++ b/adhocracy-plus/config/settings/base.py
@@ -201,6 +201,8 @@ PARLER_LANGUAGES = {
                 }
 }
 
+PARLER_ENABLE_CACHING = False
+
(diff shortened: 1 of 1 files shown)


phillimorland (approved):
Nice!

PR #846, opened 2020-10-16
organisations/assests/react_language_choice: stay on tab if removed l…

Rineee:
…anguage is not active tab

But I am still wondering why the anchor links work here?


phillimorland (approved):
I don't know why the anchors worked here, I can only think that as we use anchors a lot in Opin there's a setting somewhere that we don't have in a+? either way it's looking great!

PR #860, opened 2020-10-22
DIID Pm 2020 email logo issue

phillimorland:


adhocracy-plus/templates/email_base.html
```diff
@@ -12,7 +12,7 @@
     <tr>
             <td style="padding: 0px; position: relative; top: -4.5em;" align="center">
                     {% if organisation and organisation.logo %}
-                    <div style="width: 100px; height: 100px; box-shadow: 0 2px 18px 0 rgba(0, 0, 0, 0.08); background-color: #ffffff;">
+                    <div style="width: auto; height: 100px; box-shadow: 0 2px 18px 0 rgba(0, 0, 0, 0.08); background-color: #ffffff;">
```
rmader, line 15: Can we use that for a+ as well? In that case it would be great to split out into an own commit/MR for `master`

phillimorland: I only tested this in 2 webmails with a rectangle logo, if I was going to add it to a+ I would want to test on all webmails with square logos, is it worth it for the rebase?

rmader (approved):
Nice, LGTM

PR #861, opened 2020-10-22
SSO registration form

rmader:
Todo:
 - [x] add saml2 test server
 - [x] add overwrite-app
 - [x] set email address received from saml as primary address (if no other primary exists already)
 - [x] use verified-status of email address from saml as indicator for signup form
   - [x] redirect to saml signup form if the address in question is not yet verified
   - [x] set email as verified when signup form is submitted
   - [x] finish signup form template
   - [ ] (optionally) add middleware that force-redirects to saml signup form if user email address is not yet verified


rmader:
@Rineee @fuzzylogic2000 @phillimorland: this is getting close to be ready. If you like you can pick this up - it should now be mostly about finishing the signup form.

rmader:
I think this is good to go now - not perfect, but quite usable. There's still translations missing and the fields are still the original a+ ones - however I thought we might first land this and the translation changes and then add a follow up to adopt the form.

fuzzylogic2000 (approved):
It does read well (how do I test locally?) and I think it's a good idea to merge and then take care of the last bits that are still left!
Is there a way to tell git, that the files were moved to a different folder and changed? I always like to see the difference.

rmader:
`make saml-server` should start the saml server - apart from that it should work with `make watch` - the dev-config should be all set up.

fuzzylogic2000:
Whoop whoop, that works!

PR #862, opened 2020-10-23
Semipublic projects

Rineee:


apps/projects/rules.py
```diff
@@ -7,11 +7,17 @@
...
 
 rules.add_perm('a4_candy_projects.change_project',
                is_superuser | is_initiator)
 
 rules.add_perm('a4_candy_projects.view_project',
+               is_superuser | is_initiator | is_moderator |
+               ((is_public | is_semipublic | is_org_member | is_project_member)
+                & is_live))
+
+rules.add_perm('a4_candy_projects.participate_in_project',
```
rmader, line 20: I wonder if we should move these to a4, use `a4projects....` rules everywhere and, where adoption is needed, overwrite the `a4projects....` rules using `rules.set_rule()`

rmader: For the record: that's what we do in mB, so we should most likely do the same here.

rmader: Done

apps/projects/overwrites.py
```diff
@@ -0,0 +1,12 @@
...
+from adhocracy4.projects.enums import Access
+
+
+def overwrite_access_enum_label():
+    Access.__labels__ = {
+        Access.PRIVATE: _('Only invited users can participate (private).'),
+        Access.PUBLIC: _('All users can participate (public).'),
+        Access.SEMIPUBLIC: _('All users can view the content but only invited '
+                             'users can participate (semipublic).')
+    }
```
Rineee, line 12: The labels in the story are a little different: https://taiga.liqd.net/project/liqd-liquid-software/us/3433?milestone=239
Sorry :see_no_evil:

rmader: Adopted :)

Rineee: I am sooo sorry, but no :) Its the wording below the acceptance criteria (wording in basic settings:)..

rmader: Oh, didn't see that, thanks!

concerning review: IMO you shouldn't be sorry for that - by properly reviewing the PR you do me and all of us a big favour, saving us time and effort afterwards. Right now it's easy to fix up, costing me way less effort than having somebody to open an issue etc. So please, rather nitpick often :)

apps/dashboard/forms.py
```diff
@@ -20,12 +20,19 @@ class DashboardProjectCreateForm(ProjectCreateForm):
...
+                       'participate (public).')),
+                    (project_models.Access.SEMIPUBLIC.value,
+                     _('All users can see project tile and content, only '
+                       'invited users can participate (semi-public).')),
+                    (project_models.Access.PRIVATE.value,
+                     _('Only invited users can see project tile and content '
+                       'and can participate (private).'))
                 ]
-            )
+            ),
```
Rineee, line 37: So cant we remove the labels here then?

apps/projects/overwrites.py
Rineee: :+1:

apps/dashboard/forms.py
rmader: I'd prefer to leave this in as it's actually how it should be done - the overwrite is just a weird workaround IMO. I'd really like to get rid of it again :/

Rineee: Alright, I understand. But then maybe a note that this is actually not used for the moment? Because it might be confusing/not so transparent for someone who doesnt know the code..

rmader: Right, will add

rmader: Done

Rineee: Thank you!

Rineee (commented):
Its really cool and as clean as it can be, so lets merge!

PR #876, opened 2020-10-29
WIP makefile: add initial make command - not working

phillimorland:
i'll leave WIP on it so you can look at the folder structure


fuzzylogic2000:
Cool! I will need to think and read up before I have an idea what to do here.

fuzzylogic2000:
Ok, I think our plan is not working like we hoped it would. 
The exclude option is for languages (so, `--exclude de_DE` would leave out making the German po-file). It seemed to work for the `make po-fork`, because the makemessages just takes the first locale folder it is given in the settings and puts the translation files into that. So, it always takes the locale-fork folder here.

I'll try to write down what we want and what we have once more. And if it's only to order my thoughts.
We want to have two translation folders, one coming from our main branch and also to be used in our default project. Then, we would like to have a second one overwriting these translations only in a fork or branch.
The current problem is, that the locale paths in the settings are used for determining the order of translations, but also to find the folder to write the translations to. Ahahaha, we only add the second folder to the settings in the fork! I will try that out!

fuzzylogic2000:
Well, that did work. Kind of. But only like you already added it @phillimorland with the two folders locale-source and locale-fork. That is weird and I might try again on Monday. :woman_shrugging:

PR #884, opened 2020-11-03
renovate.json: add minor updates to automerge

phillimorland:


fuzzylogic2000 (approved):
Yeah!

rmader:
FTR: I'm not entirely sure but IIRC a minor upgrade for jQuery broke things for as at some point. So this is a bit more risky again - but still probably a good idea :)

phillimorland:
I know, we did speak about that but as we test a+ dev so much I think it's probably fine, things like that will get caught quickly, maybe less of a good idea on less tested projects though

PR #886, opened 2020-11-03
diid: Add a default_idp setting

rmader:
When set, the login directly forwards to the specified IDP, in
contrast to showing a full list. This does not prevent other IPDs
to send valid login tokens - thus the user needs to make sure the
own service is only registered on the desired IDP.


rmader:
@fuzzylogic2000 this is quite trivial, will go for it for now :)

PR #892, opened 2020-11-03
account/signup: add additional checkboxes into if statement

phillimorland:


adhocracy-plus/templates/account/signup.html
```diff
@@ -13,7 +13,7 @@ <h1>{% trans "Register" %}</h1>
         {{ form.non_field_errors }}
         {% csrf_token %}
         {% for field in form %}
-            {% if field is not form.terms_of_use and field is not form.terms_of_use_extra and field is not form.get_newsletters %}
+            {% if field is not form.terms_of_use and field is not form.terms_of_use_extra and field is not form.get_newsletters and field is not form.data_protection and field is not form.get_notifications %}
```
rmader, line 16: Hehe, I somewhat wonder if we exclude more than include, actually :)

rmader (approved):
LGTM, thx!

PR #895, opened 2020-11-04
Ka 2020 11 translations diid

fuzzylogic2000:
I think, we can now add all missing for DIID to the locale-fork folder and we should be fine!


Rineee (approved):
Nice!

PR #896, opened 2020-11-04
Makefile: Merge po-fork back into po

rmader:
Appently they behave mutually exclusive anyway - if `po-fork` should be
used, `po` produces broken output.

@fuzzylogic2000 does that look sensible to you?


phillimorland:
@rmader this won't work will it?  it will still cause conflicts? point was that we don't touch locale source in fork and vice versa or am I missing something?

fuzzylogic2000 (approved):
Good idea! But can't we even merge that into master? Or would that break when the locale-fork folder isn't there?

fuzzylogic2000 (approved):
yeah!

PR #903, opened 2020-11-04
Makefile: Only call sed if file actually exists

rmader:
`sed` will complain when feeded with an empty file list. Thus use a
loop to call each file individually.

Sorry, didn't test the previous solution before.


phillimorland (approved):
super cool! :)

PR #910, opened 2020-11-05
docs: add transifex documentation

phillimorland:
I think this is right, @Rineee please feel free to correct me?!


Rineee (approved):
That is great, thank you @phillimorland for documenting this!

PR #913, opened 2020-11-06
Makefile: add folders to ignore to makemessages to avoid double entri…

Rineee:
…es of code lines

The double code line references for adhocracy4 entries come once from node_modules/adhocracy4 and once from venv/.../adhocracy4

So for django.po, it now ignores the ones from node_modules and for djangojs.po it now ignores the ones from venv. 

What do you think?

I had to use the $CURDIR because it somehow only works with absolute paths...


phillimorland (commented):
So I think this is super cool! but when I run make po it deletes loads of translations but I don't know if this is my usual issue with my computer deleting translations?

fuzzylogic2000 (approved):
I think all removed strings are also added elsewhere!

PR #914, opened 2020-11-06
Ks 2020 11 pulled translations from transifex

Rineee:
Pulled translations from transifex, then made po again to get the spaces right and deleted commented out strings from source files.

I didnt run the new makemessages (see [ks-2020-11-add-ignore-to-makemessages](913)), because I wasnt sure if we want that. So line references are now still double, but could then be removed by running the new make po.


fuzzylogic2000 (approved):
Yeah!

PR #933, opened 2020-11-10
apps/organisations: make module still be shown when organisation does…

fuzzylogic2000:
… not exist
fixes https://sentry.liqd.net/sentry/aplus-prod/issues/1092/ and https://sentry.liqd.net/sentry/aplus-dev/issues/1197/

This happens whenever a module view is accessed with a wrong organisation slug in the url. As a solution, the organisation info is now not shown. I think a 404 would be better, but don't know how to achieve that and think this is fine, as it very seldomly happens or shouldn't happen. (On dev it's quite clear what happened, but on prod, it's a bit weird as the organisation just doesn't exist.)

### DIFF
--- a/adhocracy-plus/templates/header_lower.html
+++ b/adhocracy-plus/templates/header_lower.html
@@ -5,12 +5,12 @@
     <div class="col-md-10 offset-md-1">
         <nav class="header-lower__nav">
 
-            {% if ORGANISATION.logo %}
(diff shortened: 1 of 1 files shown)


rmader:
Hm, why is it possible to access the module at all in this case?

rmader (approved):
Nice, LGTM

Issue #930, opened 2020-11-10
DIID - missing translation / wording changes


CarolingerSeilchenspringer:
**URL:** registration via shibboleth
**user:** 
**expected behaviour:** German text
**behaviour:** english text including teilhabe.diid.hhu.de: 'You are about to use your HHU account to login to teilhabe.diid.hhu.de. As a final step, please complete the following form' 
**important screensize:**
**device & browser:** 
**Comment/Question:** please change to the following:
English: You are about to use your HHU account to login. As a final step, please complete the following form.
German: Sie sind dabei, sich mit Ihrem HHU-Konto einzuloggen. Als letzten Schritt füllen Sie bitte das folgende Formular aus.

Screenshot?

PR #935, opened 2020-11-11
djangosamle2_overwrits/signup: update wording and add translation fix…

phillimorland:
…es #930

still a few translation deletions but I think it doesn't matter as it will take them from the source-po files?


fuzzylogic2000 (approved):
Yeah!
There are a two strings lost in the djangojs.po files, BUt that doesn't matter here, as we will still get the translations from the locale-source. It is weird, though!

PR #936, opened 2020-11-11
templates/header_lower & footer_upper: always use organisation from view

fuzzylogic2000:
A better solution for #933 and also including the footer_lower

But should go into the release after the next one, to be tested on dev for a while!

### DIFF
--- a/adhocracy-plus/templates/header_lower.html
+++ b/adhocracy-plus/templates/header_lower.html
@@ -5,36 +5,34 @@
     <div class="col-md-10 offset-md-1">
         <nav class="header-lower__nav">
 
-            {% if ORGANISATION and ORGANISATION.logo %}
-              {% url 'organisation' organisation_slug=ORGANISATION.slug as organisation_url %}
-                <a href="{% url 'organisation' organisation_slug=ORGANISATION.slug %}" class="header-lower__logo-box" rel="home">
-                    <img src="{% thumbnail ORGANISATION.logo '0x160' %}" height="80" alt="{{ ORGANISATION.name }}" class="header-lower__logo-img" />
+            {% if organisation.logo %}
+              {% url 'organisation' organisation_slug=organisation.slug as organisation_url %}
+                <a href="{% url 'organisation' organisation_slug=organisation.slug %}" class="header-lower__logo-box" rel="home">
+                    <img src="{% thumbnail organisation.logo '0x160' %}" height="80" alt="{{ organisation.name }}" class="header-lower__logo-img" />
                 </a>
-            {% elif ORGANISATION %}
+            {% else %}
               <div class="header-lower__brand u-bg-body">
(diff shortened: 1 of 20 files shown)


adhocracy-plus/templates/header_lower.html
```diff
@@ -5,36 +5,34 @@
...
-              {% url 'organisation' organisation_slug=ORGANISATION.slug as organisation_url %}
-                <a href="{% url 'organisation' organisation_slug=ORGANISATION.slug %}" class="header-lower__logo-box" rel="home">
-                    <img src="{% thumbnail ORGANISATION.logo '0x160' %}" height="80" alt="{{ ORGANISATION.name }}" class="header-lower__logo-img" />
+            {% if organisation.logo %}
+              {% url 'organisation' organisation_slug=organisation.slug as organisation_url %}
+                <a href="{% url 'organisation' organisation_slug=organisation.slug %}" class="header-lower__logo-box" rel="home">
+                    <img src="{% thumbnail organisation.logo '0x160' %}" height="80" alt="{{ organisation.name }}" class="header-lower__logo-img" />
                 </a>
-            {% elif ORGANISATION %}
+            {% else %}
```
Rineee, line 13: I get an error if the logo is not there, because the slug is missing.. probably only because of the capital letters?

fuzzylogic2000: Hmm, no, I just changed that back from the fix before (where the organisation just isn't shown, when it's wrong in the url). What exactly is the error?

fuzzylogic2000: Oh, or maybe in some views it does need the correct organisation slug from the url? Which view was that in?

Rineee: The error is `Reverse for 'organisation' with keyword arguments '{'organisation_slug': ''}' not found. 1 pattern(s) tried: ['(?P<organisation_slug>[-\\w_]+)/$']` in the header_lower.html 

But sorry, I commented in the wrong line, it is line 15 I meant ..

adhocracy-plus/templates/header_lower.html
```diff
@@ -5,36 +5,34 @@
...
               <div class="header-lower__brand u-bg-body">
                 <a class="header-lower__brand--link font-weight-bold" href="{% url 'organisation' organisation_slug=ORGANISATION.slug %}">
```
Rineee, line 15: so here it should be changed to organisation_slug=organisation.slug, shouldnt it?

fuzzylogic2000: Oh my, I looked at the wrong line. Yes, exactly, that slipped back in in a rebase! Will fix it! Thank you!

Rineee (approved):
Excellent, I like that!

PR #937, opened 2020-11-11
apps/organisations: make sure published poll in unpublished project d…

fuzzylogic2000:
…oesn't break the organisation_tags

another thing that shouldn't happen, but did: if a project has a poll that is not finished added, the preview breaks: https://aplus-stage.liqd.net/liqd/dashboard/projects/this-is-a-poll/basic/

fixes https://sentry.liqd.net/sentry/aplus-prod/issues/1001/events/latest/ and https://sentry.liqd.net/sentry/aplus-stage/issues/1090/?query=is%3Aunresolved


phillimorland (approved):
think I tested correctly?! looks good though!

PR #939, opened 2020-11-11
Pm 2020 11 design fixes

phillimorland:
yo, design fixes along with the logo update, if that one looks funky once it'd on dev there's a second option in Projekte/adhocracy.plus/Kreation/Redesign/201021-birthday-campaign/201021-birthday-campaign-jpg/2x


fuzzylogic2000 (commented):
Yes! Awesome! 
It just readds the thing that breaks the project and module sites when the organisation slug is empty (https://sentry.liqd.net/sentry/aplus-prod/issues/1092/). I will fix that the easy way now and then also add that to the proper fix for later.

PR #940, opened 2020-11-12
translations: Remove country postfixes

rmader:
We don't make use of them atm (only having one for each language) but
could lead to making us fall back to untranslated string. Most
importantly, users using `en` would not see overwrites from `en_GB`.

@fuzzylogic2000: do we need to change anything for transifex here?


fuzzylogic2000 (approved):
Looks good!
I don't know about transifex, though. The languages there are with postfixes. I don't know how it will handle the change. I'll try it before merge!

PR #941, opened 2020-11-12
translations: Show languages in their own language

rmader:
It's likely that someone looking to change the language doesn't know how the own language is called on another language. Always show languages in their own language in the language switcher.

image

Depends on and includes #940


fuzzylogic2000 (approved):
It works!
But I also pushed these strings to transifex and they are shown in the translated English version. We'll see what happens there.

PR #942, opened 2020-11-12
adds/djangosaml2_overwrites: Add a middleware to guard saml signup

rmader:
djangosaml2 does not offer a way to not login the user before doing
extra validation steps. In order to force users to sign up before
being able to use the website, add a simple middleware, redirecting
them to the signup as long as their email is not verified.

Logout and wagtail pages are excepted from these rules.


fuzzylogic2000 (approved):
Yeah totally works!

PR #943, opened 2020-11-13
cms: make sure only existing page-types can be choosen

fuzzylogic2000:
fixes https://sentry.liqd.net/sentry/aplus-stage/issues/999/


Rineee (approved):
Yes! Finally understood how it works :) 
I first tried to add a FormPage to Root and then select that for an important page, but that obviously doesnt work, because it doesnt have an url. Maybe we should also think about making that impossible?

PR #945, opened 2020-11-16
organisations/sitemaps: fixup for using the new access enum

Rineee:
fixes https://sentry.liqd.net/sentry/aplus-prod/issues/1256/


fuzzylogic2000 (approved):
Yeah!

PR #962, opened 2020-11-19
apps: Merge questions and likes to interactiveevents

rmader:
And lots of refactoring, including:
- move some template stuff back from the general project/module templates to the interactive event template. This includes the livestream, which thus looks a bit better
- move and rename lots of js code. Also add a lookup to webpack to make js code from apps behave more like from npm packages
- rename questions to livequestions, so we have the same naming as in mB (and we always know what exactly were talking about)

Includes https://github.com/liqd/adhocracy-plus/pull/977 - unfortunately without proper commits this time :(


phillimorland (changes requested):
Currently you cannot like as a logged out user

rmader:
> Currently you cannot like as a logged out user

Smells like something we should have a test for :)

phillimorland (commented):
not in keeping with design but you can ask Marcelo about that and it's failing tests so won't merge

rmader:
Ah right, same issue as on https://github.com/liqd/adhocracy-plus/pull/977 - with the new template we need to adopt the test

rmader:
FTR, it comes down to

image

vs

rmader:
Ok, reverted that change as my goal was not to mess with designs

PR #964, opened 2020-11-23
header: change logo back to normal

phillimorland:


fuzzylogic2000 (commented):
Cool! But we should also delete the image!

phillimorland:
I thought just leave it for next celebration as will make it easier?

fuzzylogic2000 (approved):

PR #965, opened 2020-11-23
Pm 2020 11 module detail clean

phillimorland:
I have checked the 4 views I can think of an this doesn't effect them but I wonder if i am missing something, views checked;
multimodule
timeline module
single module
interactive event module


adhocracy-plus/templates/a4modules/module_detail.html
```diff
@@ -53,7 +53,7 @@
     {% endblock %}
 
     {% block content %}
-        <div class="project-header{% if project.image %} project-header--image{% endif %} row" style="{% if view.project.image %}
+        <!-- <div class="project-header{% if project.image %} project-header--image{% endif %} row" style="{% if view.project.image %}
```
phillimorland, line 56: @fuzzylogic2000 do we really not need any of this or have I forgotten something super obvious?

PR #977, opened 2020-11-24
Pm 2020 11 module detail clean

phillimorland:
only way to make it work with designs was to add another template

### DIFF
--- a/adhocracy-plus/templates/a4modules/module_detail.html
+++ b/adhocracy-plus/templates/a4modules/module_detail.html
@@ -59,21 +59,14 @@
             <div class="container">
                 <div class="row mb-5">
                     <div class="col-md-8 col-lg-6 offset-md-2 offset-lg-3">
-                        {% with container=project.containers.first %}
+                        <!-- {% with container=project.containers.first %}
                             {% if container %}
                                 <a href="{{ container.get_absolute_url }}" class="project-header__link" data-embed-target="external">
                                     <i class="fa fa-arrow-left" aria-hidden="true"></i>
                                     {% blocktrans with name=container.name %}to project collection {{ name }}{% endblocktrans %}
                                 </a>
                             {% endif %}
-                        {% endwith %}
-                        <!-- Live stream -->
-                        {% for phase in module.phases %}
-                          {% if phase == module.active_phase and live_stream %}
-                              {{ live_stream.live_stream | transform_collapsibles | richtext }}
-                              <span class="label label--active">•{% trans ' Live now' %}</span>
-                          {% endif %}
-                        {% endfor %}
+                        {% endwith %} -->
 
                         <h1 class="project-header__title">{{ project.name }}</h1>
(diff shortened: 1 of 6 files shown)


fuzzylogic2000 (commented):
It's a good idea and looks good!
But currently, the button is there and I can go to the new view when the phase is not active. What do we currently do about that? We have some info on the page when it starts or that participation is not possible at the moment? Maybe we can merge this and then add some info/behaviour for that next?

phillimorland:
@fuzzylogic2000 shall I just hide the button when phase not active?

fuzzylogic2000:
@phillimorland That wouldn't work for after the phase, because people are still meant to view the results, right? Maybe don't display the react before the phase and instead The sentence "participation is not possible at the moment, it starts at..."

adhocracy-plus/templates/a4modules/module_detail.html
```diff
@@ -59,22 +48,6 @@
             <div class="container">
                 <div class="row mb-5">
                     <div class="col-md-8 col-lg-6 offset-md-2 offset-lg-3">
-                        {% with container=project.containers.first %}
-                            {% if container %}
-                                <a href="{{ container.get_absolute_url }}" class="project-header__link" data-embed-target="external">
-                                    <i class="fa fa-arrow-left" aria-hidden="true"></i>
-                                    {% blocktrans with name=container.name %}to project collection {{ name }}{% endblocktrans %}
```
rmader, line 66: Do we want to remove that for all modules?

apps/questions/templates/a4_candy_questions/question_module_detail.html
```diff
@@ -17,9 +13,10 @@ <h3>{{ module.name }}</h3>
                 <div class="mb-3">
                     {{ module.description | transform_collapsibles | richtext }}
                 </div>
-                <a class="btn btn--primary" href="{% url 'module-detail' module.project.organisation.slug module.slug %}">{% trans 'Go to event' %}</a>
+                {% if not module.future_phases %}
```
phillimorland, line 16: @fuzzylogic2000 this seemed like a better choice as we don't repeat ourselves now

adhocracy-plus/templates/a4modules/module_detail.html
phillimorland: I didn't find any trace of project containers being used as it seemed like a left over from mB so thought I would clean it up

apps/questions/templates/a4_candy_questions/question_module_detail.html
fuzzylogic2000: Yes, you're right!

adhocracy-plus/templates/a4modules/module_detail.html
rmader: Ah right

rmader:
Changes included in https://github.com/liqd/adhocracy-plus/pull/962

PR #978, opened 2020-11-24
header_lower: add additional header outside positioned div for mobile…

### LINKED ISSUE #771: Org page mobile
info and overview wrong on mobile

phillimorland:
… to ensure can be clearly seen fixes #771

It's a little messy but that nav is very tricky on small screens with the positioning so this was easiest was to solve it


fuzzylogic2000 (approved):
I think it's a good solution with the include!

PR #991, opened 2020-12-02
contrib/management: Add command to clean up unverified users

rmader:
Strictly speaking we should check for the email verified status here,
but if a user never logged in that pretty much implies the same.

Prints a list of deleted users, which we'll recieve as mail from
a cron job.

### DIFF
--- a/apps/contrib/management/commands/cleanup_unverified_users.py
+++ b/apps/contrib/management/commands/cleanup_unverified_users.py
@@ -1,0 +1,39 @@
+from datetime import timedelta
+
+from django.core.management.base import BaseCommand
+from django.utils import timezone
+
+from apps.users.models import User
+
+
+class Command(BaseCommand):
+    help = 'Remove all users that registered more than n days ago but never ' \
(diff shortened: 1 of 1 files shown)


fuzzylogic2000 (commented):
Yes, even though never logged in and not verified should be the same in most cases (only if the admin of a platform created users, that never logged in, it might be different), I think if we delete the users that never logged in, we should also say that in the helptext and the print statement. Otherwise it might lead to confusion at a later point.
(And we could and should port this over at some point: https://github.com/liqd/a4-meinberlin/blob/66736d7060036dd4df52ae61344146096a5c4826/meinberlin/apps/newsletters/emails.py#L30 )

rmader:
@fuzzylogic2000 updated the helptext. I guess we don't need to print that for every user. Main point to not use the `verified` state is also that I'd fear to confuse something about users with several mail addresses (social login). So this should be the safer way.

I also added another parameter that would only print the output which I'd use for the cronjob at the beginning.

Mind having another look?

fuzzylogic2000 (approved):
Nice!

PR #992, opened 2020-12-03
Ks 2020 11 wording changes for livequestion categories, char count and anchor

Rineee:


adhocracy-plus/assets/scss/components/_live_questions.scss
```diff
@@ -4,3 +4,7 @@
     margin-top: -0.75*$spacer;
     margin-bottom: $spacer;
 }
+
+.live_questions__text {
+    overflow-wrap: break-word;
```
fuzzylogic2000, line 9: Nice! I think else where we use "word-break: break-word", but this one seems to be better (compatible to earlier browser versions)!

fuzzylogic2000 (approved):
Very nice and because of the well split commits easy to read!

PR #994, opened 2020-12-07
NOT FOR REVIEW: apps/ideas: add new phase for add idea, comment and rate, add blueprint

phillimorland:
@fuzzylogic2000 Nice task! :) I wonder if it might be a little easy just because the collect phase in ideas (the last one) is the same, unless I miss something? but still it gets familiarity with connection to dashboard and phases.
Oh and I would maybe define it's. branch of a+ not a4, that might be me just being dim though :p


fuzzylogic2000:
Coolio!

Issue #749, opened 2020-08-17
Idea tiles and submit idea button


phillimorland:
Carry over issue from DIID but should be for whole platform 
04 – Regarding the idea tiles + Submit Idea Button, I did the whole screen cause it was easier to explain:
https://app.zeplin.io/project/5f05dd0342a6025af2786627/screen/5f3408482228711c4b444cff/
Screenshot?

phillimorland:
@mcastro-lqd ok looking into this, the list of ideas are not tiles, they are just list items. do we want list items and project/subject tiles to all look the same?

phillimorland:
@mcastro-lqd Here is what these lists cureently look like in the different modules:
https://aplus-dev.liqd.net/teststadt/projects/idea-challenge-with-map/?mode=list

mcastro-lqd:
> @mcastro-lqd Here is what these lists currently look like at moment in couple different modules, adding the shadow from design looks a bit weird with the background colour, should I remove the background colour from our list elements? I just want to check before doing it as it effects all modules with lists.
> https://aplus-dev.liqd.net/teststadt/projects/idea-challenge-with-map/?mode=list
> https://aplus-dev.liqd.net/teststadt/projects/lorem-ipsum/

Yes, I'm actually already making new designs using white background instead cause since we had to make aplus legal and separate the organisation content from our content using the grey background, I wouldn't use it inside the organisation content.

fuzzylogic2000:
@phillimorland @mcastro-lqd What's the status? Is it done? Should it be part of a story/redesign? Or what exactly is still missing?

mcastro-lqd:
@fuzzylogic2000 I think its kind of done, from what I got we are changing little by little each time we work on improving a  module

fuzzylogic2000:
Nice! As we already remember to improve in small steps, I will close this issue!

PR #995, opened 2020-12-07
assets: fix btn allignment and design issues

### LINKED ISSUE #1013: #3506 interactive event
**URL:** Interactive event Modules like https://aplus-dev.liqd.net/teststadt/interactiveevents/interactive-event-3/
**important screensize:** *Both*
**device & browser:** Safari *Version 14.0 (15610.1.28.1.9, 15610)*
**Comment/Question:**

Sorry, I think I tested too early cause I didn't see some of these new things and I can't find my old issue anymore to add these things, so I just created this new one instead.
 
[m-201208-interactive-event-questions.pdf](https://github.com/liqd/adhocracy-plus/files/5659526/m-201208-interactive-event-questions.pdf)
[d-201208-interactive-event-questions.pdf](https://github.com/liqd/adhocracy-plus/files/5659523/d-201208-interactive-event-questions.pdf)

phillimorland:
partial fix for #749
fixes #1013

### DIFF
--- a/adhocracy-plus/assets/js/app.js
+++ b/adhocracy-plus/assets/js/app.js
@@ -1,5 +1,6 @@
 import 'bootstrap' // load bootstrap components
 import 'django'
+import 'select2' // used to skin select element used in livequestions
 import 'slick-carousel' // for project timeline
 
 import './unload_warning.js'
@@ -53,6 +54,10 @@ function init () {
     slidesToShow: 1,
     slidesToScroll: 1
   })
+
+  if ($.fn.select2) {
+    $('.js-select2').select2()
+  }
 }
 
 document.addEventListener('DOMContentLoaded', init, false)
--- a/apps/projects/templates/a4_candy_projects/project_detail.html
+++ b/apps/projects/templates/a4_candy_projects/project_detail.html
@@ -247,7 +247,7 @@
                 </div>
                 {% endif %}
                 <!-- these blocks are only filled when in phase view -->
-                <div class="col-md-8 col-lg-6 offset-md-2 offset-lg-3">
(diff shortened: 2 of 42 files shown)


adhocracy-plus/assets/js/app.js
```diff
@@ -53,6 +54,10 @@ function init () {
     slidesToShow: 1,
     slidesToScroll: 1
   })
+
+  if ($.fn.select2) {
+    $('.js-select2').select2()
+  }
 }
```
phillimorland, line 61: I'm leaving this in until we know which we'll use but I'm not calling it on the select in this pull request

rmader: :+1:

apps/projects/templates/a4_candy_projects/project_detail.html
```diff
@@ -21,6 +21,17 @@
 {% endif %}
 {% endblock %}
 
+<!-- ensure cookie overlay for embedded videos -->
+{% block extra_css %}
+  {{ block.super }}
+    <link rel="stylesheet" href="{% static 'dsgvo_video_embed.css' %}" />
```
rmader, line 27: Can you remove it from `livequestion_module_detail.html` again then? Otherwise we load it twice.

rmader (approved):
One nit, otherwise from a quick test here looks good!

apps/projects/templates/a4_candy_projects/project_detail.html
phillimorland: oh good point, done.

rmader: Thanks!

PR #1017, opened 2020-12-08
apps/interactive_events: Restart polling interval when adding question

rmader:
Directly fire a update request and reset the timer, so the own question
is shown faster.


Rineee (approved):
Supercool, much better now! Lets also do this in meinBerlin!

PR #1018, opened 2020-12-09
interactiveevents: reaaranging responsive design of filters so no mov…

phillimorland:
…ement
That was very annoying, don't like design so much but movement has stopped


fuzzylogic2000 (approved):
Cool!

PR #1019, opened 2020-12-09
Ks 2020 12 move livequestion category select to own component

Rineee:
Tried to style it, but didnt really succeed.. @phillimorland I think there is padding missing inside the select box, but couldnt figure out how to do it, could only always add padding around the box :grimacing:


fuzzylogic2000 (approved):
Nice!

PR #1021, opened 2020-12-10
interactiveevent: additional padding to select arrow and add star to …

phillimorland:
…checkbox

### DIFF
--- a/adhocracy-plus/assets/scss/components/_select2.scss
+++ b/adhocracy-plus/assets/scss/components/_select2.scss
@@ -30,6 +30,10 @@
         }
     }
 
+    .select2-selection--single .select2-selection__rendered {
+        padding-right: 2*$spacer;
+    }
+
     &.select2-container--open {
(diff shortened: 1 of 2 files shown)


apps/interactiveevents/templatetags/react_interactiveevents.py
```diff
@@ -27,7 +27,7 @@ def react_interactiveevents(context, obj):
 
     private_policy_label = str(_('I confirm that I have read and accepted the '
                                  '{}terms of use{} and the {}data protection '
-                                 'policy{}.'))
+                                 'policy{}.*'))
```
fuzzylogic2000, line 30: Can we move it out of the tranlated string? Or is that too annoying with our special react-link-thing? Then, we should leave it for next time, when we get new trsnalations. Sorry!

fuzzylogic2000: Haha, translation is a very hard word

fuzzylogic2000 (approved):

PR #1026, opened 2020-12-10
interactiveevent: replace class with classname and undo changes and r…

### LINKED ISSUE #1023: stage: interactive event statistics issue
**URL:** https://aplus-stage.liqd.net/liqd/interactiveevents/interactive-event/
**user:** initiator/admin/moderator
**expected behaviour:** the question goes back to unanswered 
**behaviour:** by clicking on the check mark in the statistics, nothing happens and the questions disappear.
**important screensize:**
**device & browser:** 
**Comment/Question:** I am unsure what was pokered here. An alternative would be that nothing happens and the answered questions remain as they're

Bildschirmfoto 2020-12-10 um 14 12 28

phillimorland:
…eplace unsupported function

@rmader tag me back if you get something cleaner, I just wasn't sure what you were on about in your message

fixes #1023

### DIFF
--- a/apps/interactiveevents/assets/js/a4_candy_interactive_events/livequestion_box.jsx
+++ b/apps/interactiveevents/assets/js/a4_candy_interactive_events/livequestion_box.jsx
@@ -217,7 +217,7 @@ export default class QuestionBox extends React.Component {
           aria-hidden="false"
         >
           <div className="container">
-            <div class="row">
+            <div className="row">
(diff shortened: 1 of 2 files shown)


rmader (approved):
LGTM! For the record: the reason why we duplicate the the properties in the state is to allow us to already remove the question from the list until the next update happens - via polling, which currently only happens every 5 seconds. An alternative to this would be to directly request an update as we do in the question form - however, it might be even better to have a local state in case a device has a slow internet connection.

PR #1039, opened 2020-12-15
apps/projects: only show private info for modules, not events

fuzzylogic2000:
like https://github.com/liqd/a4-meinberlin/pull/3331


fuzzylogic2000:
@Rineee @phillimorland @rmader I totally forgot about this one. Can someone merge?

Rineee (commented):
Yes!

PR #1073, opened 2021-01-20
apps/organisations/models: do not consider draft modules for sorting …

Rineee:
…of project list

Stumbled across this when looking at the project sorting.. since we were also considering draft modules, the sorting was messed up, e.g. projects with a non-published module with 'active' phase were considered active even though the only published module ended already and thus were shown among active projects.

What do you think? I would then also port this over to kosmo..


fuzzylogic2000 (approved):
Yes! Very nice! Good catch!
This was obviously already there when we changed the modules to being draft.

PR #1093, opened 2021-02-04
apps/embed: Some 3rd-party cookie fixes

rmader:
The cookie test was completely broken, not sure why that should have worked in the past. And a workaround for https://code.djangoproject.com/ticket/31933


phillimorland (approved):
super awesome cool! :)

PR #1107, opened 2021-02-15
Ks 2021 02 update js

Rineee:
So I tried to find all that needed adjustment when updating A4 after the js refactoring.

I initially wanted to port over [this PR](https://github.com/liqd/a4-meinberlin/pull/3386/files) from mB, but realized that we dont need that here, as we dont have plans and the other map initialization is done in A4, right?

Also not sure about [this PR]([url](https://github.com/liqd/a4-meinberlin/pull/3387/files))? The jQuery wasnt there before and I didnt add it, because polygons seem to work fine..

And finally I am not sure, why the credits dont show for the maps.. they do locally, because they are set explicitly [here](https://github.com/liqd/adhocracy-plus/blob/23611c276e5f3a1ba1ad5b4429359e006b6014b7/adhocracy-plus/config/settings/base.py#L494), but this must be overwritten on the servers somewhere, because there they dont show. Maybe we could have a look together tomorrow @rmader ?

### DIFF
--- a/adhocracy-plus/assets/js/app.js
+++ b/adhocracy-plus/assets/js/app.js
@@ -13,7 +13,7 @@ import '../../../apps/newsletters/assets/dynamic_fields.js'
 // expose react components
 import {
   comments as ReactComments,
-  comments_async as ReactCommentsAsync,
+  commentsAsync as ReactCommentsAsync,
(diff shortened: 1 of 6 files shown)


rmader (approved):
LGTM!

PR #1114, opened 2021-02-16
organisations/sitemaps: Return 404 if organisation does not exist

rmader:
This avoids 500 errors when search engines rescan projects of deleted
organisations.

Closes https://sentry.liqd.net/organizations/liqd/issues/1272/


fuzzylogic2000 (approved):
Nice! That's where that error came from!

PR #1126, opened 2021-02-24
apps/notifications: Handle non-existing organisatiov in project delete

rmader:
Fixes https://github.com/liqd/adhocracy-plus/issues/1124

Note: if the organization was deleted, there won't be any email send - as there are no initiators any more


fuzzylogic2000 (approved):
Yes, that does work!

PR #1127, opened 2021-02-25
Kl 2021 02 captcha

fuzzylogic2000:
This should be enough to start working on the DB - we can add the more complicated wagtail form captcha stuff later.


rmader (approved):
LGTM. Nit: I personally prefer if commits within a MR do not first add and then remove something again, even though it's understandable in the case here :)

fuzzylogic2000:
@rmader  If you prefer, that I squash it, I can also do that.

fuzzylogic2000:
@rmader It did work! Do you want to merge or should I?

rmader (approved):
LGTM!

PR #1128, opened 2021-02-25
webpack/js improvements

rmader:
This is mostly a port of https://github.com/liqd/a4-meinberlin/pull/3440 and https://github.com/liqd/a4-meinberlin/pull/3445

It reduces the js size considerably, hopefully making browsing on slow devices faster.


fuzzylogic2000 (approved):
Reads well! Let's see if it also works. :)

PR #1137, opened 2021-03-02
cms/wagtail: add new fields for wagtails AbstractFormField

### LINKED ISSUE #1129: wagtail update broke contact form
**URL:** https://aplus-dev.liqd.net/ (at the bottom)
**user:** 
**expected behaviour:** 
**behaviour:** 
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Was caused by
https://github.com/liqd/adhocracy-plus/blob/3064be108030cd56178dcb56a45d0fb7808209c6/apps/cms/contacts/migrations/0002_new_wagtail_fields.py
because of this commit in wagtail:
https://github.com/wagtail/wagtail/commit/d0bb803645bc60fd56584339f71a8ff8415f45ea#

fuzzylogic2000:
fixes #1129

But to not make the migration run without the changes, we might need to change the history of master :grimacing: @rmader Can we have a look together?

### DIFF
--- a/apps/cms/contacts/models.py
+++ b/apps/cms/contacts/models.py
@@ -115,30 +115,35 @@ class FormPage(AbstractEmailForm):
         fields = list(super().get_form_fields())
         fields.insert(0, FormField(
             label='receive_copy',
+            clean_name='receive_copy',
             field_type='checkbox',
             help_text=_('I want to receive a copy of my message'),
             required=False))
 
         fields.insert(0, FormField(
             label='message',
+            clean_name='message',
             help_text=_('Your message'),
             field_type='multiline',
             required=True))
 
         fields.insert(0, FormField(
             label='email',
+            clean_name='email',
             help_text=_('Your email address'),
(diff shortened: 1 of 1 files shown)


fuzzylogic2000:
@rmader No, it's actually fine and also shows the content of the fields again when the migration is run first and this added later! :relieved:

rmader (approved):

rmader:
@fuzzylogic2000 do we need to port this over to mB? mb-dev currently doesn't appear to have a feedback site (we should fix that).

fuzzylogic2000:
@rmader I thought I checked it. I think the update is already in the release and it is fine, because we do the from differently there (you have to create it in wagtail, it's not automatically added)

fuzzylogic2000:
@rmader To be very sure, I checked it all again. :) It is in the release and fine.
And it is fine, because we don't add the FormField-stuff into the fields like that, so we also didn't need the extra migration.

PR #1141, opened 2021-03-03
apps/embed/assets: Work around login popup not closing

rmader:
This affects cases when people login from the login-link shown
e.g. above comments.


rmader:
Taking the freedom to merge, see https://github.com/liqd/a4-meinberlin/pull/3463

PR #1142, opened 2021-03-04
Kl 2021 03 contact captcha

fuzzylogic2000:


fuzzylogic2000:
To be checked: is the captcha send with the email?

rmader (approved):
Approach makes sense to me, LGTM!

PR #1143, opened 2021-03-04
add ky and ru languages story #3875

phillimorland:
…languages, add nl, ky, ru as cms simple page languages, add new translation class for legal pages story #3875


adhocracy-plus/config/settings/base.py
```diff
@@ -200,12 +200,14 @@
...
-    ('nl', _('Dutch'))
+    ('nl', _('Dutch')),
+    ('ky', _('Kyrgyz')),
+    ('ru', _('Russian'))
 ]
 
 PARLER_LANGUAGES = {
     1: [{'code': language_code} for language_code, language in LANGUAGES],
-    'default': {'fallbacks': ['en', 'de'],
+    'default': {'fallbacks': ['en', 'de', 'nl,' 'ky', 'ru'],
```
phillimorland, line 210: they are not needed here right?

phillimorland:
I'm not sure if theres more to it then this until we have the transifix? if so just let me know, oh and I wasn't sure on the paler languages so left a question.

adhocracy-plus/config/settings/base.py
fuzzylogic2000: Yes, I think so, for the organsiation languages.

fuzzylogic2000 (approved):
Yeah!

PR #1160, opened 2021-03-10
Pm 2021 03 language tests

phillimorland:


fuzzylogic2000:
No, I think, this is fine, we can't test every new language, I would say?!

PR #1170, opened 2021-03-16
docs: update transifex procedure

Rineee:
I tried pushing new strings without pulling translations first and translations get not lost, to I updated the docs


phillimorland (approved):
looking good!

PR #1173, opened 2021-03-17
Ks 2021 03 undo language switch hack

Rineee:
Translations agaaain :)

This is to undo https://github.com/liqd/adhocracy-plus/pull/941

I first updated the english source files manually, then pushed that to transifex, translated there and pulled again.


phillimorland (approved):
looking good!

PR #1177, opened 2021-03-22
Ks 2021 03 validate that username is not already in use as email

Rineee:
This should fix 
https://sentry.liqd.net/organizations/liqd/issues/1353/?project=32&query=is%3Aunresolved 
and also
https://sentry.liqd.net/organizations/liqd/issues/1358/?project=3&referrer=alert_email
if we port it over to meinBeriln


Rineee:
Just realized that I forgot to make the strings translatable, will do that tomorrow, but if you have time, maybe you can already check if it works..

phillimorland:
@Rineee this still lets me save my email as me username, thats not meant to happen right?

Rineee:
> @Rineee this still lets me save my email as me username, thats not meant to happen right?

Yes, I think that is fine, it its the same account. You can then still login, right? The problem before was that it let you save an email address as username that was already used as email address for another account which led to not being able to login in anymore.

I will fix this PR now, it seems that husky precommit doesnt work?

phillimorland (approved):
Nice!

PR #1186, opened 2021-03-24
templates/500.html: adding inline css for styling 500 page

phillimorland:


phillimorland:
@fuzzylogic2000 that was a lie, I found it in admin, all looks nice now

Rineee (approved):
Yes!

PR #1199, opened 2021-04-01
makefile: silence non-critical warning when probing for gsed

goapunk:
Fix the `make: command: Commant not found` warning during build. Adding the semicolon forces bash to skip some builtin optimizations and actually execute the command.


phillimorland:
@goapunk awesome! sorry took a while to get to this, when you push a pull request, tag any/all of dev team as reviwers then we'll get email reminder so we don't forget :p

PR #1216, opened 2021-04-16
Kl 2021 04 new strings

fuzzylogic2000:
New po files as pushed to transifex and few translations to German pulled from transifex

### DIFF
--- a/locale-source/locale/en/LC_MESSAGES/django.po
+++ b/locale-source/locale/en/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: adhocracy-plus\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-03-17 17:57+0100\n"
+"POT-Creation-Date: 2021-04-16 10:47+0200\n"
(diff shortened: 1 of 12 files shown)


locale-source/locale/en/LC_MESSAGES/django.po
```diff
@@ -4580,3 +4581,54 @@ msgstr "Login with social media account"
 #: adhocracy-plus/templates/socialaccount/snippets/provider_list.html:23
 msgid "Login with "
 msgstr "Login with "
+
```
Rineee, line 4584: Did you leave these in on purpose, so translators can copy from here to the new strings with adhocracy+ replaced? Im asking bc unfortunately transifex also shows commented out strings, so normally we should remove them, so translators dont have extra work with it...

fuzzylogic2000: Yes, exactly! And I already copied from them for German. We have the other thing in the docs, so will also remember to delete them next time. :)

Rineee: Ha, perfect!

fuzzylogic2000:
Ah, I don't get why travis is failing. Problem for Monday...

fuzzylogic2000:
Whoop whoop! Fixed itself!

PR #1242, opened 2021-04-30
WIP Ks 2021 04 remove old questions and likes apps

Rineee:
DO NOT MERGE :skull_and_crossbones: 
until a4_candy_questions and a4_candy_likes tables are all dropped (also in forks -> DIID and KOSMO)
Need to keep the apps for that, so we can add a migrations for dropping the tables.


phillimorland:
@Rineee is this still waiting for the dropped tables?

Rineee:
Oh right! I think we just forgot this and the tables are still there.... @fuzzylogic2000 Do you remember what the plan was? Should we just delete them?

fuzzylogic2000:
@Rineee I can remember that we talked about it, but now I am confused why the models are still there, when the tables are dropped. So, we have to find out again and then check at least KOSMO as well as the current release of a+. Or maybe we should also wait until we know what will happen to DIID?

PR #1243, opened 2021-05-03
Ks 2021 04 add image dashboard component to interactiveevent

Rineee:
This is just adding the event image to the model and view (can be accessed with extra_fields.event_image), still needs the adding and styling in template.

@kha So the template where the event image should be added is [this one](https://github.com/liqd/adhocracy-plus/blob/master/apps/interactiveevents/templates/a4_candy_interactive_events/module_detail.htmll). There you can acces the image by extra_fields.event_image, because I put it in the context of InteractiveEventModuleDetail view, so in [this file](https://github.com/liqd/adhocracy-plus/blob/master/apps/interactiveevents/views.py) I hope this helps!


khamui:
Working fine on my local as well. Merging into master so i can build the frontend on top this.

khamui:
> This is just adding the event image to the model and view (can be accessed with extra_fields.event_image), still needs the adding and styling in template.
> 
> @Kha So the template where the event image should be added is [this one](https://github.com/liqd/adhocracy-plus/blob/master/apps/interactiveevents/templates/a4_candy_interactive_events/module_detail.htmll). There you can acces the image by extra_fields.event_image, because I put it in the context of InteractiveEventModuleDetail view, so in [this file](https://github.com/liqd/adhocracy-plus/blob/master/apps/interactiveevents/views.py) I hope this helps!

the first link seems to not working properly. Anyways with the second link i see the template file. Thank you!

PR #1250, opened 2021-05-04
module/interactiveevent: showing optional event image

khamui:
Using the new module property extra_fields.event_image to show
an event image if existing. Re-layouting detail screen (also
for mobile).
WIP: function of 'read more' button to be discussed.

### DIFF
--- a/adhocracy-plus/assets/scss/components/_button.scss
+++ b/adhocracy-plus/assets/scss/components/_button.scss
@@ -175,6 +175,26 @@
         color: $text-color;
         cursor: not-allowed;
     }
+
+    &.collapsible {
+        .link--less {
+            display: none;
+        }
+
+        .link--more {
+            display: block;
+        }
+    }
+
+    .show + &.collapsible {
+        .link--less {
+            display: block;
+        }
+
+        .link--more {
+            display: none;
+        }
+    }
 }
 
 .btn--none {
--- a/adhocracy-plus/assets/scss/components/_label.scss
+++ b/adhocracy-plus/assets/scss/components/_label.scss
@@ -32,8 +32,9 @@
 
 .label--live {
     background-color: $body-bg;
-    border: 2px solid $brand-danger;
+    border: 1px solid $brand-danger;
     color: $brand-danger;
+    text-transform: uppercase;
 }
 
 .label--CONSIDERATION {
--- a/apps/interactiveevents/templates/a4_candy_interactive_events/module_detail.html
+++ b/apps/interactiveevents/templates/a4_candy_interactive_events/module_detail.html
@@ -1,18 +1,62 @@
 {% extends extends %}
-{% load i18n discovery_tags ckeditor_tags wagtailcore_tags %}
+{% load i18n discovery_tags ckeditor_tags wagtailcore_tags contrib_tags %}
 
 {% block phase_content %}
-    <div class="bg--light">
-        <div class="container">
-            <div class="col-lg-10 offset-lg-1 u-bg-body">
-                <h3>{{ module.name }}</h3>
-                <div class="alert" role="alert">
-                    <i class="fas fa-info-circle" aria-hidden="true"></i>
+    <div class="container">
+        <div class="col-12">
+            <div class="alert" role="alert">
+                <i class="fas fa-info-circle" aria-hidden="true"></i>
                 {% trans 'During this event you can hand in questions with your smartphone' %}
+            </div>
+        </div>
+        <div class="row mx-md-4 my-4 justify-content-center">
+            {% if extra_fields.event_image %}
+                <div class="col-md-6 mb-4">
+                    <img src="/media/{{ extra_fields.event_image }}" />
                 </div>
-                <div class="mb-3">
-                    {{ module.description | transform_collapsibles | richtext }}
-                </div>
+            {% endif %}
+            <div class="col-md-6">
+                {% for phase in module.phases %}
+                    {% if phase == module.active_phase and extra_fields.live_stream %}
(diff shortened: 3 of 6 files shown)


khamui (commented):
with the latest commit the 'read more' / 'read less' should also work (thanks @phillimorland !!!)

adhocracy-plus/assets/scss/_layout.scss
```diff
@@ -7,6 +7,32 @@
...
+        &.left {
+            margin: 2rem 1rem 2rem 5rem;
+        }
+
+        &.right {
+            margin: 2rem 5rem 2rem 1rem;
+        }
+    }
+}
+
```
phillimorland, line 35: have you considered using pre-existing bootstrap here, like cols? https://getbootstrap.com/docs/4.1/layout/grid/ all designs are based on 12 col structure so it will fit for every screen size

adhocracy-plus/assets/scss/_variables.scss
```diff
@@ -10,7 +10,7 @@ $brand-tertiary: #008dc5 !default; // prussian blue
 $brand-success: #a3ef90 !default;
 $brand-info: $brand-primary !default; // $brand-primary from platform.scss
 $brand-warning: #ffc107 !default;
-$brand-danger: #a72b1e !default;
+$brand-danger: #f00 !default;
```
phillimorland, line 14: This variable is used for many elements, this effects all error messages and as we use lighten() for lots of them, changing it means that colour dissapears. If we need a new colour we can either define a new variable in this file or in the css file for whichever component we are working on, or I sometimes check with the designer if they think we should change it everywhere as it's best not to add lots of colours to a site. here I would leave it as is and check with design after it's on dev so they can see it.

adhocracy-plus/assets/scss/components/_button.scss
```diff
@@ -175,6 +175,26 @@
...
+    .show + &.collapsible {
+        .link--less {
+            display: block;
+        }
+
+        .link--more {
+            display: none;
+        }
+    }
 }
```
phillimorland, line 198: I would add the `transition: none` here as it looks a little confusing when swapping between 2

adhocracy-plus/assets/scss/components/_label.scss
```diff
@@ -4,7 +4,8 @@
     color: contrast-color($text-color);
     font-size: $font-size-xs;
     font-weight: normal;
-    border-radius: 0.4em;
+    border-radius: 0.8em;
+    border: 1px solid $text-color;
     padding: 0.2em 0.7em;
 }
```
phillimorland, line 11: again we use `label` on many elements so updating needs to be something thats considered carefully, checking the component library can help when changing these things to see if it has unintended effects, this will also help you see the danger colour change issues. I would just apply what we need to the `live--label` class. Border radius should be something thats the same for all labels so we can change it here or check with designer, I normally go with whichever is easier or I think looks better :p

apps/interactiveevents/views.py
```diff
@@ -21,6 +21,7 @@ def get_context_data(self, **kwargs):
         context['extra_fields'] = \
             models.ExtraFieldsInteractiveEvent.objects.filter(
                 module=self.module).first()
+        context['phase_desc_limit'] = 75
         return context
```
phillimorland, line 25: do we need this as well if we have the truncate?

phillimorland (changes requested):
looking good! just a few things to alter that are more to do with our processes then anything :)

adhocracy-plus/assets/scss/_layout.scss
khamui: goodstuff, thiiink i can achieve the same with pre-existing bootstrap classes.

apps/interactiveevents/views.py
khamui: ah good catch. leftovers from before

apps/interactiveevents/templates/a4_candy_interactive_events/module_detail.html
```diff
@@ -1,18 +1,62 @@
...
+                                    </div>
+                                    <button
+                                        class="btn--link collapsible    "
+                                        type="button" data-toggle="collapse"
+                                        data-target="#collapsibleContent"
+                                        aria-expanded="false"
+                                        aria-controls="collapseExample"
+                                    >
+                                        <span class="link--more">{% trans 'Read more' %}</span>
+                                        <span class="link--less">{% trans 'Read Less' %}</span>
```
phillimorland, line 51: these links are both visible at the same time on firefox, not sure why given the styling, maybe incorrect nesting

apps/interactiveevents/templates/a4_candy_interactive_events/module_detail.html
```diff
@@ -1,18 +1,62 @@
...
-                <div class="alert" role="alert">
-                    <i class="fas fa-info-circle" aria-hidden="true"></i>
+    <div class="container">
+        <div class="col-12">
+            <div class="alert" role="alert">
+                <i class="fas fa-info-circle" aria-hidden="true"></i>
                 {% trans 'During this event you can hand in questions with your smartphone' %}
+            </div>
+        </div>
+        <div class="row mx-md-4 my-4 justify-content-center">
```
phillimorland, line 12: x-axis margins on rows shouldn't be needed as all designs are based on 12 cols so if it lookas too wide, then it probably needs to be less columns

phillimorland (changes requested):
again looking good but few bootstrap procedures to look at

phillimorland (approved):
awesome! :)

PR #1257, opened 2021-05-06
#4131 changing tabs for live event and styling

phillimorland:


phillimorland:
This is still a little odd on mobile but I want to talk to Marcelo about it and easiest if he can see on dev so hoping it can be merged and then I can talk to him, I also need to do some a11y tool tips I think but for a different pull request, will add tasks to taiga

fuzzylogic2000 (approved):
LGTM!

PR #1261, opened 2021-05-07
Ks 2021 05 adding tests

Rineee:
There is much more to test, maybe for Monday :)

### DIFF
--- a/tests/dashboard/test_dashboard_views.py
+++ b/tests/dashboard/test_dashboard_views.py
@@ -1,0 +1,102 @@
+import pytest
+from django.urls import reverse
+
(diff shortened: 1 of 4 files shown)


fuzzylogic2000:
You assigned us, but I only get alerts right away when you request a review! I will still review it, though! :)

fuzzylogic2000 (commented):
Very nice! And even though, github doesn't show it yet: coveralls says +1.2% !! :partying_face:

Rineee:
> You assigned us, but I only get alerts right away when you request a review! I will still review it, though! :)

Ups :see_no_evil:

PR #1265, opened 2021-05-10
apps/account: move regsiter info text above button - fixes #387

### LINKED ISSUE #387: a+ text under register button not visible before click - usertesting
**behaviour:** after capcha is added to reg form, information below registration button is not visible without scrolling
**important screensize:** smaller then desktop

fuzzylogic2000:
Luckily it still looks fine with the paragraph in the form. I don't know if it's alright for a11y, though?! @phillimorland Do you know?

### DIFF
--- a/adhocracy-plus/templates/account/signup.html
+++ b/adhocracy-plus/templates/account/signup.html
@@ -54,17 +54,17 @@
 
         {% include 'a4_candy_contrib/includes/form_field.html' with field=form.captcha %}
 
-        <div class="u-spacer-bottom-double">
-            <button class="btn btn--primary" type="submit">{% trans "Register" %}</button>
-        </div>
-    </form>
-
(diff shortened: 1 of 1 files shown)


phillimorland (approved):
LGTM

phillimorland:
@fuzzylogic2000 which part do you mean isn't right for a11y?

fuzzylogic2000:
> @fuzzylogic2000 which part do you mean isn't right for a11y?

@phillimorland I meant if it was alright to just put the paragraph into the form, because it might not make a lot of sense not being a label or helptext. But I guess as it is in the flow of the html, it's alright?

phillimorland:
@fuzzylogic2000 it's still within the flow and it has no associated input so it would be incorrect for it to be a label so I would say go with it!

PR #1267, opened 2021-05-10
module/interactiveevents: image path

khamui:
depends on: https://github.com/liqd/adhocracy4/pull/720

### DIFF
--- a/adhocracy-plus/assets/scss/components/_live_questions.scss
+++ b/adhocracy-plus/assets/scss/components/_live_questions.scss
@@ -85,3 +85,7 @@
     position: relative;
     bottom: -7*$spacer;
 }
+
+.livequestion-detail__image {
+    width: 100%;
+}
--- a/apps/interactiveevents/templates/a4_candy_interactive_events/module_detail.html
+++ b/apps/interactiveevents/templates/a4_candy_interactive_events/module_detail.html
@@ -1,5 +1,5 @@
 {% extends extends %}
-{% load i18n discovery_tags ckeditor_tags wagtailcore_tags contrib_tags %}
+{% load i18n discovery_tags ckeditor_tags wagtailcore_tags contrib_tags thumbnail %}
 
 {% block phase_content %}
     <div class="container">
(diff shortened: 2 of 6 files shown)


adhocracy-plus/templates/a4modules/module_detail.html
```diff
@@ -90,25 +90,6 @@ <h1 class="project-header__title">{{ project.name }}</h1>
                 </div>
             {% endif %}
             {% endblock %}
-            <div class="row">
```
khamui, line 93: not sure if this a4module is meant to be modified?

apps/interactiveevents/templates/a4_candy_interactive_events/module_detail.html
```diff
@@ -14,14 +14,22 @@
...
             {% endif %}
             <div class="col-md-5 my-2">
                 {% for phase in module.phases %}
                     {% if phase == module.active_phase and extra_fields.live_stream %}
                         <span class="label label--live">•{% trans ' Live now' %}</span>
                         &nbsp;
+                    {% elif not phase == module.active_phase %}
+                        <span class="label label--info">
+                            {% if not module.past_phases or module.future_phases %}
+                                In {{ week_countdown }} weeks
```
fuzzylogic2000, line 28: We do have a property from a4, we use in the module tiles. I will find it!
Also, if you wanted to do it like that, you still need to make the string translatable. That's done with the blocktrans tag, where you can also add variables to the text.

apps/interactiveevents/views.py
```diff
@@ -21,8 +24,19 @@ def get_context_data(self, **kwargs):
...
             models.ExtraFieldsInteractiveEvent.objects.filter(
                 module=self.module).first()
+
+        if self.module.future_phases:
+            startdate = self.module.future_phases.first().start_date
+            context['week_countdown'] = self.is_future_event(startdate)
+
         return context
 
+    def is_future_event(self, start_date):
```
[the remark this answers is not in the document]
fuzzylogic2000: Yes, if you want to use it in only one template, it's good to add it to the context of the view! (But here, we already have something to do it. Or was it weeks in the design? I think, we only go up to days.)

adhocracy-plus/assets/scss/components/_live_questions.scss
```diff
@@ -85,3 +85,7 @@
     position: relative;
     bottom: -7*$spacer;
 }
+
+.livequestion__detail--image {
```
fuzzylogic2000, line 89: Sorry, I didn't see that yesterday, but the image is not a variant. A variant needs to be used together with the thing it's a variant from like `btn` and `btn-primary`. Öhm, I need to look BEM up again, should it be `livequestion__detail__image`?

fuzzylogic2000 (commented):
Very nice!
I only have one more very small thing, but feel so sorry about finding that only now, that I will just quickly fix it!

adhocracy-plus/assets/scss/components/_live_questions.scss
fuzzylogic2000: Ok, I decided that livequestion-detail is the block as livequestion__detail wasn't yet used. And image is the element, so it's `livequestion-detail__image` now.

khamui: yes, makes sense, the **B**lock is `livequestion-detail` and the **E**lement is `image`. thanks for changing it :) and sorry that i paid less attention to the naming here.

PR #1283, opened 2021-05-11
apps/cms/home_page: update hero text to include char limit, update fi…

### LINKED ISSUE #1233: char limit landing page title and subtitle
to prevent stuff like: https://github.com/liqd/a4-kosmo/issues/119

phillimorland:
…eld names fixes #1233


fuzzylogic2000 (changes requested):
This is cool! But the migration didn't catch the rename. To make it do a rename, you can rename first with a migration and then change the other stuff with a second migration.

fuzzylogic2000 (approved):
Nice! It doesn't even just cut the fields off, but only tells you with the next save that it's too long. Which is great if there is some too long content somewhere.
What is a bit weird is that the 150 chars for the richtextfield are not really that (probably because the markup also has to be saved and counts toward the char count). Should we make it a bit longer for that reason or is 150 still fine? (At least our prod server has less. :p)
Screenshot from 2021-05-12 13-07-15

phillimorland:
@fuzzylogic2000 right so we really shouldn't be able to add paragraphs or line breaks with as that also breaks design, I didn't pick up on that when testing, will spend little more time but this might be a dud :(

phillimorland:
@fuzzylogic2000 ok you can't remove <p> from richtext so I would say leave it at 150, hopefully will discourage spaces!

fuzzylogic2000:
Alright! :tada:

PR #1286, opened 2021-05-12
apps/interactiveevents: read more event description and phase descrip…

khamui:
…tion separately (variant A)

### DIFF
--- a/apps/interactiveevents/templates/a4_candy_interactive_events/module_detail.html
+++ b/apps/interactiveevents/templates/a4_candy_interactive_events/module_detail.html
@@ -33,7 +33,26 @@
                 <span class="label">{% trans 'Interactive' %}</span>
                 <h1>{{ module.name }}</h1>
                 <div class="mb-2">
-                    {{ module.description | transform_collapsibles | richtext }}
+                    {% if module.description|length > 180 %}
+                        <div id="collapsibleContentEvent" class="collapse show u-no-transition">
+                            {{ module.description|truncatechars:160|richtext }}
+                        </div>
+                        <div id="collapsibleContentEvent" class="collapse u-no-transition">
+                            {{ module.description|richtext }}
+                        </div>
+                        <button
+                            class="btn--link collapsible    "
+                            type="button" data-toggle="collapse"
+                            data-target="#collapsibleContentEvent"
+                            aria-expanded="false"
+                            aria-controls="collapseExample"
+                        >
+                            <span class="link--more">{% trans 'Read more' %}</span>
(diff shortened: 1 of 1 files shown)


khamui:
variantA

apps/interactiveevents/templates/a4_candy_interactive_events/module_detail.html
```diff
@@ -33,7 +33,26 @@
                 <span class="label">{% trans 'Interactive' %}</span>
                 <h1>{{ module.name }}</h1>
                 <div class="mb-2">
-                    {{ module.description | transform_collapsibles | richtext }}
+                    {% if module.description|length > 180 %}
+                        <div id="collapsibleContentEvent" class="collapse show u-no-transition">
+                            {{ module.description|truncatechars:160|richtext }}
+                        </div>
```
phillimorland, line 39: you use 180 and the 160 on collapsible elements, doens't that mean that 20 are cut off?

apps/interactiveevents/templates/a4_candy_interactive_events/module_detail.html
```diff
@@ -45,17 +64,17 @@ <h1>{{ module.name }}</h1>
...
-                        {% if module.phases.first.description|length > 80 %}
-                            <div id="collapsibleContent" class="collapse show u-no-transition">
-                                {{ module.phases.first.description|truncatechars:80|richtext }}
+                        {% if module.phases.first.description|length > 180 %}
+                            <div id="collapsibleContentPhase" class="collapse show u-no-transition">
+                                {{ module.phases.first.description|truncatechars:160|richtext }}
                             </div>
-                            <div id="collapsibleContent" class="collapse u-no-transition">
+                            <div id="collapsibleContentPhase" class="collapse u-no-transition">
                                 {{ module.phases.first.description|richtext }}
```
phillimorland, line 72: nice naming!

phillimorland (changes requested):

apps/interactiveevents/templates/a4_candy_interactive_events/module_detail.html
khamui: so if description is bigger than 180, it will show 160 + read more
if description is smaller than 180, it will show full length.

so in case of 181 or 212, it will show 160 + read more
trying to avoid if someone uses 181 chars, then behind read more is only one letter.

or am I wrong? yeah its a bit of knot in the head..

fuzzylogic2000 (approved):
LGTM

PR #1288, opened 2021-05-12
jd 2021 05 upgrade to new sentry sdk

goapunk:
The raven sdk has been deprecated for a while now, we should switch to the new sdk. This requires changes in our server setup as well.


goapunk:
I've added the sentry test route for testing purpose, we should probably drop it before we merge this.

goapunk:
@phillimorland I rebased and bumped it to the latest version. Needs to be merged together with https://github.com/liqd/admin/pull/335

PR #1289, opened 2021-05-12
interactiveevents:livequestion_box: ensure checked status is correct …

### LINKED ISSUE #1281: #4131 interactive event design of tabs: statistics button outline issue
**URL:** https://aplus-dev.liqd.net/teststadt-luca/interactiveevents/interaktive-veranstaltung-13/
**user:** all
**expected behaviour:** The statistics button is not outlined in blue when going back from "Statistics" to the questions and only remains blue as long as you see the statistics
**behaviour:** Apparently, the outline of the button remains blue when clicking on the "cross"in the statistics and you get back to the questions
**important screensize:** 
**device & browser:** all
**Comment/Question:** Is the cross really necessary? Intuitively speaking I would rather click on the statistics button again if I want to get back to the questions. If it's not possible to change the outline this could be an option maybe.

Bildschirmfoto 2021-05-11 um 10 51 24
Bildschirmfoto 2021-05-11 um 10 51 33

phillimorland:
…event when using close button fixes #1281 and small spacing fix

### DIFF
--- a/apps/interactiveevents/assets/js/a4_candy_interactive_events/livequestion_box.jsx
+++ b/apps/interactiveevents/assets/js/a4_candy_interactive_events/livequestion_box.jsx
@@ -222,8 +222,8 @@ export default class QuestionBox extends React.Component {
                           {textDisplayOnScreen}
                         </a>
                       </div>}
-                    <div className="livequestions__action-bar--btn">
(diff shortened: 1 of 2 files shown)


fuzzylogic2000 (approved):
Looking good!

PR #1290, opened 2021-05-14
apps/organisations/landing page: show different text for no projects …

### LINKED ISSUE #311: organisation landing page: text for emtpty list does not take into account if user is logged in
The text is: 
Currently, there are no public participation processes. Sign in here (LINK) or read more in About (LINK).
There is a check if there is actually an about-tab and that is only displayed, when it exists. When we also check if the user is logged in, we have four versions of the text. That seems a bit over the top. Any ideas, @hmkay

fuzzylogic2000:
…when user is logged in - fixes #311

I think when this is merged, we can also push the strings to transifex!
Or maybe only after the a4-update (after the merge of the PR there) in which we can also tag a4!

### DIFF
--- a/apps/organisations/templates/a4_candy_organisations/organisation_landing_page.html
+++ b/apps/organisations/templates/a4_candy_organisations/organisation_landing_page.html
@@ -84,7 +84,9 @@
                         </ul>
                     {% endif %}
                 {% else %}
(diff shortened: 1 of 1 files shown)


Rineee (approved):
Yes, cool!

PR #1291, opened 2021-05-14
Kl 2021 05 double phase content

### LINKED ISSUE #1263: [a+] #4129 manual testing: Interactive Event adapting design of module detail view
**URL:** https://aplus-dev.liqd.net/frede-test/projects/test-projekt-4129/
**user:** -
**expected behaviour:** 
**behaviour:** 
**important screensize:**
**device & browser:** tested on a mac with safari and chrome+ iPhone with safari
**Comment/Question:** 

1. Text lengths: On the detail page there are two text: The first one is the short description from the basic settings and is shown in full length. The second one is the description from the phases and it is shorter then the first one but gets shortened. Is this how it was meant? @mcastro-lqd 
Bildschirmfoto 2021-05-09 um 22 53 30

2. The "live"-tag was missing, when the event was in the active phase.

3. The event-image was not visible / broken:
Bildschirmfoto 2021-05-09 um 23 12 40

fuzzylogic2000:
I don't think, we should treat only that one module type so differently by checking for the phases... I added a FIXME to the commit message and will also add an issue.
fixes #1263 (the last thing to fix: 4.)

### DIFF
--- a/adhocracy-plus/templates/a4modules/module_detail.html
+++ b/adhocracy-plus/templates/a4modules/module_detail.html
@@ -103,13 +103,16 @@
                         </ul>
                     </nav>
                     {% block module_description %}
+                    {% if module.phases.first.type != 'a4_candy_interactive_events:issue' %}
                     <h1 class="project-header__title">{{ module.name }}</h1>
                     <p class="project-header__description">{{ module.description }}</p>
+                    {% endif %}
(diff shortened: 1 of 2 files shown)


Rineee (approved):
Yes, cool! Also that you added the fixme and issue

PR #1294, opened 2021-05-14
Ks 2021 05 more tests

Rineee:
@fuzzylogic2000 I was tempted to remove [this](https://github.com/liqd/adhocracy-plus/blob/3d521b9307dd415c76cbc6fcd0ee8fdb7a7e3de3/apps/dashboard/views.py#L192-L197) from ModuleDeleteView, because I think we dont need it anymore, right? Because the dropdown with deletion option only shows if module is draft? But then I saw that these lines were only added a year ago, so not sure anymore.. do we still need that?

### DIFF
--- a/tests/dashboard/test_dashboard_views.py
+++ b/tests/dashboard/test_dashboard_views.py
@@ -1,4 +1,5 @@
 import pytest
+from django.contrib.messages import get_messages
 from django.urls import reverse
 
 from adhocracy4.modules.models import Module
@@ -81,6 +82,9 @@ def test_module_create(client, project):
     response = client.post(url)
 
     assert redirect_target(response) == 'dashboard-module_basic-edit'
+    messages = list(get_messages(response.wsgi_request))
+    assert len(messages) == 1
+    assert str(messages[0]) == 'The module was created'
 
     response2 = client.post(response.url)
     assert not response2.context_data['form'].is_valid()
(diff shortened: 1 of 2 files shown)


fuzzylogic2000:
> @fuzzylogic2000 I was tempted to remove [this](https://github.com/liqd/adhocracy-plus/blob/3d521b9307dd415c76cbc6fcd0ee8fdb7a7e3de3/apps/dashboard/views.py#L192-L197) from ModuleDeleteView, because I think we dont need it anymore, right? Because the dropdown with deletion option only shows if module is draft? But then I saw that these lines were only added a year ago, so not sure anymore.. do we still need that?

I think we do, because it is two different things to me: we only show the button when the module is a draft anymore. But someone could still post to the view when a module is already published. Or add a button somewhere that does that. So, the view should still handle that!

fuzzylogic2000 (approved):
Sooooo cool! You made it to > 85%! Nearly +1%! :tada:

fuzzylogic2000:
Is it still WIP? Or can we merge?

Rineee:
> > @fuzzylogic2000 I was tempted to remove [this](https://github.com/liqd/adhocracy-plus/blob/3d521b9307dd415c76cbc6fcd0ee8fdb7a7e3de3/apps/dashboard/views.py#L192-L197) from ModuleDeleteView, because I think we dont need it anymore, right? Because the dropdown with deletion option only shows if module is draft? But then I saw that these lines were only added a year ago, so not sure anymore.. do we still need that?
> 
> I think we do, because it is two different things to me: we only show the button when the module is a draft anymore. But someone could still post to the view when a module is already published. Or add a button somewhere that does that. So, the view should still handle that!

Alright, then I will also test it. In meinBerlin, we dont have that btw, I'll add an issue.

Rineee:
> Is it still WIP? Or can we merge?

Sorry, just quickly added testing if module is draft when deleting. Now its ready!

PR #1295, opened 2021-05-14
WIP: Kl 2021 05 pm block clean

fuzzylogic2000:
trying to replace #494 
Migrations in wagtail are harder than I thought. I am following https://www.caktusgroup.com/blog/2019/09/12/wagtail-data-migrations/, but am only half through the data migration. With changing the columns, I still need to find out what to copy where.


fuzzylogic2000:
> @fuzzylogic2000 whats the status here? should we keep it open?

@phillimorland Hmm, not sure actually. It would be nice to have, but was surprisingly much work. Let's keep it open until the next a+ sprint and close when we don't do it then! I will add it to amilestone.

PR #1296, opened 2021-05-14
Ks 2021 05 disable unload warning outside dashboard

### LINKED ISSUE #1226: Browser says I haven´t sent information
**URL:** https://kosmo-dev.liqd.net/liqd/subjects/2021-00017/
**user:** admin
**expected behaviour:** I reported a comment and then just go via the "back" button back to the discussion
**behaviour:** My browser warns me that I haven´t sent the information that I put in - but I did!
**important screensize:** Desktop
**device & browser:** MacBook, Mozilla
**Comment/Question:** Is this something in our hands or is it something with the Browser I use?

Bildschirmfoto 2021-04-26 um 11 30 02

Rineee:
fixes #1226

Although this is not the proper way, I guess we should find out why the browser thinks the information in Report Modal is not sent, even though it is..

### DIFF
--- a/adhocracy-plus/assets/js/app.js
+++ b/adhocracy-plus/assets/js/app.js
@@ -3,7 +3,6 @@ import 'django'
 import 'select2' // used to skin select element used in livequestions
 import 'slick-carousel' // for project timeline
 
-import './unload_warning.js'
 import '../../../apps/actions/assets/timestamps.js'
 import '../../../apps/dashboard/assets/ajax_modal.js'
 import '../../../apps/maps/assets/map-address.js'
--- a/adhocracy-plus/templates/a4dashboard/base_dashboard.html
+++ b/adhocracy-plus/templates/a4dashboard/base_dashboard.html
@@ -1,5 +1,5 @@
 {% extends "base.html" %}
-{% load i18n rules thumbnail %}
+{% load i18n rules thumbnail static %}
 
 {% block content %}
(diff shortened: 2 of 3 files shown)


adhocracy-plus/templates/a4dashboard/base_dashboard_project.html
```diff
@@ -138,3 +138,8 @@
         </div>
     </div>
 {% endblock %}
+
+{% block extra_js %}
```
fuzzylogic2000, line 142: Do we also have a dashboard_base? Or will we also get problems when we use that?

fuzzylogic2000 (commented):
Looking good! Just one question.

adhocracy-plus/templates/a4dashboard/base_dashboard_project.html
Rineee: Yes, we do have a dashboard_base. I just did it the way its done in meinBerlin, there its added to base_dashboard_project. But I actually dont think we get problems if I add it to dashboard base. So should I do that instead?

fuzzylogic2000: As you like! You can just merge or move it to dashboard_base. Well, I think, I prefer the latter! :)

fuzzylogic2000 (approved):
:unlock:

PR #1310, opened 2021-05-20
Pm 2021 05 webpack cofig update

phillimorland:


phillimorland:
This will hopefully work when on dev, it's mostly clean up, not so much optermisation but might still be a touch faster. I'm not sure if the comments are too much though?

fuzzylogic2000 (approved):
Very nice! I think, there can't be too many comments, I love them!
But let's wait with merge until after release to thoroughly test before the next release.

khamui:
nice!

PR #1317, opened 2021-05-25
ideas/serializers: add creator and created to serializer

Rineee:
@khamui  Can you work with the format of the date?

### DIFF
--- a/apps/ideas/api.py
+++ b/apps/ideas/api.py
@@ -23,6 +23,9 @@ class IdeaViewSet(ModuleMixin,
 
     def get_queryset(self):
         ideas = Idea.objects\
-            .filter(module=self.module)\
+            .filter(module=self.module) \
+            .annotate_comment_count() \
+            .annotate_positive_rating_count() \
+            .annotate_negative_rating_count() \
             .order_by('created')
         return ideas
--- a/apps/ideas/serializers.py
+++ b/apps/ideas/serializers.py
@@ -15,7 +15,7 @@ class IdeaSerializer(serializers.ModelSerializer):
     class Meta:
         model = Idea
         fields = ('pk', 'name', 'description', 'creator', 'created',
-                  'comment_count', 'positive_rating_count',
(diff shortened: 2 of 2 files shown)


fuzzylogic2000 (approved):
Very nice! Should we just merge and fix the date later if need be?

Kha:
> @Kha Can you work with the format of the date?

I can't, because I'm not @khamui

Rineee:
> > @Kha Can you work with the format of the date?
> 
> I can't, because I'm not @khamui

:see_no_evil: :sweat_smile: Ups, sorry and thanks for the hint!

khamui:
@Rineee so i tested this endpoint. The date is fine and works!

Two things i figured out, one is that if its possible to include the `Ref. No.` information?
And the `description` and maybe `title` field are containing html tags. Is it possible to just send the string without html tags?

For example: `"<p>wowowowowow</p>"` without the `<p>`

fuzzylogic2000:
> Two things i figured out, one is that if its possible to include the `Ref. No.` information?
> And the `description` and maybe `title` field are containing html tags. Is it possible to just send the string without html tags?
> 
> For example: `"<p>wowowowowow</p>"` without the `<p>`

Oh no, right, we have richtext. The question is: are we allowed to get rid of the markup or do we also need to display it in the app (the <p> doesn't matter, but we also have bold and italic and maybe more?).

khamui:
Syntactically I think JSX in react-native does not understand any html notations. There are third party libs which make it possible, but I would suggest to not create that dependency, because....

To meet and being flexible in terms of designs i think we should only pass plain text. So the responsibility of its appearance is wherever it is used. In this case the app.

Rineee:
@khamui I added the reference_number and as discussed left rich text until we have a solution/decision for that

fuzzylogic2000 (approved):
Looks good! Can we merge @khamui ? Then, we can use it in the idea views!

PR #1319, opened 2021-05-31
Ks 2021 05 add app accessible projects api

Rineee:
Just tell me if we need more fields or I added unnecessary stuff :)

Available under /api/app-projects

### DIFF
--- a/apps/projects/serializers.py
+++ b/apps/projects/serializers.py
@@ -1,0 +1,28 @@
+from django.utils.html import strip_tags
+from rest_framework import serializers
+
+from adhocracy4.projects.models import Project
+
+
+class AppProjectSerializer(serializers.ModelSerializer):
+
+    information = serializers.SerializerMethodField()
+    result = serializers.SerializerMethodField()
+    # todo: remove many=True once AppProjects are restricted to single module
+    published_modules = serializers.PrimaryKeyRelatedField(read_only=True,
+                                                           many=True)
+    access = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Project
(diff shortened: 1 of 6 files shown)


apps/projects/serializers.py
```diff
@@ -0,0 +1,28 @@
...
+
+from adhocracy4.projects.models import Project
+
+
+class AppProjectSerializer(serializers.ModelSerializer):
+
+    information = serializers.SerializerMethodField()
+    result = serializers.SerializerMethodField()
+    # todo: remove many=True once AppProjects are restricted to single module
+    published_modules = serializers.PrimaryKeyRelatedField(read_only=True,
```
fuzzylogic2000, line 12: we could also just take the first published module for now. Hmm, or it might be easier to do that in the app and when we enable multi-module projects later, the app already works with a list.

fuzzylogic2000 (approved):
Very nice! 
Should we also add some info about the state (active, past, future) directly to the project? It's easier to find out what to display in the app, then. Obviously, we can also start like this and then go on from here!

apps/projects/serializers.py
Rineee: I think (as long as its possible to add several) I prefer returning all modules that are there, it feels strange to just take one of them .. and then we can see in the app what we do about it?

Rineee:
> Very nice!
> Should we also add some info about the state (active, past, future) directly to the project? It's easier to find out what to display in the app, then. Obviously, we can also start like this and then go on from here!

Sure, we can do that.. so you mean if project is active, past or future or related to the modules?

PR #1333, opened 2021-06-02
Ks 2021 06 use rest auth token for api

Rineee:
Needs make install to run migrations.

To generate a token (or get an existing one), do
`curl -X POST http://localhost:8004/api/login/ -H 'Accept: application/json;' -d 'username=$username&password=$password'`
You can use username or email as username. The api should give you back a token, that you can also see in django admin. Then use this token to post
`curl -X POST http://localhost:8004/api/modules/$moduleId/ideas/ -H 'Accept: application/json;' -H 'Authorization: Token $token' -d 'name=my name&description=my description'`

### DIFF
--- a/adhocracy-plus/config/settings/base.py
+++ b/adhocracy-plus/config/settings/base.py
@@ -29,6 +29,7 @@ INSTALLED_APPS = (
 
     'widget_tweaks',
     'rest_framework',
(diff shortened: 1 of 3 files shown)


fuzzylogic2000 (approved):
Yeah! Works like a charm!

PR #1334, opened 2021-06-04
apps/projects: add api for module and phase info to be used in app

fuzzylogic2000:
I didn't do it with different Module serializers for different module types after all. And I hope, it all makes sense. :)


fuzzylogic2000:
@khamui It's `http://localhost:8004/api/app-modules/` for all modules from app-projects and `http://localhost:8004/api/app-modules/<module_pk>/` for a single one (and to use in the baseAPI of the app).

khamui:
very nice!

PR #1360, opened 2021-07-05
apps/exports: use export mixins from a4

### LINKED ISSUE #1359: tests: use PhaseFactory and helpers from a4
**URL:** 
**user:** 
**expected behaviour:** 
**behaviour:** 
**important screensize:**
**device & browser:** 
**Comment/Question:** after updating a4, like here https://github.com/liqd/a4-meinberlin/pull/3683

Screenshot?

fuzzylogic2000:
still very untested, just not to forget what to do.

depends on a4, which needs to be updated.

maybe also fix #1359

### DIFF
--- a/apps/budgeting/exports.py
+++ b/apps/budgeting/exports.py
@@ -3,25 +3,24 @@ from django.utils.translation import ugettext as _
 from rules.contrib.views import PermissionRequiredMixin
 
 from adhocracy4.comments.models import Comment
-from adhocracy4.exports import mixins as a4_export_mixins
+from adhocracy4.exports import mixins
 from adhocracy4.exports import views as a4_export_views
-from apps.exports import mixins as export_mixins
 
 from . import models
 
 
 class ProposalExportView(PermissionRequiredMixin,
-                         export_mixins.ItemExportWithReferenceNumberMixin,
-                         a4_export_mixins.ItemExportWithLinkMixin,
-                         a4_export_mixins.ExportModelFieldsMixin,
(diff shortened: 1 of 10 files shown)


sabinammm:
> still very untested, just not to forget what to do.
> 
> depends on a4, which needs to be updated.
> 
> maybe also fix #1359

I tested this with the most current a4 dependency (https://github.com/liqd/adhocracy4/commit/bd3d5beed4cf037157fa9be3fa1f8df832497eee) and the export works perfectly. 
BUT, the drop-downs here are broken, and the issue that seems to be fixed in master is still broken here: https://github.com/liqd/adhocracy-plus/issues/1361. I only mention in because it's unclear to me how and when it was fixed. I noticed it because the exported links to the comments lead to the project page. But rebasing and merging should fix all these issues, right @fuzzylogic2000 ?

fuzzylogic2000:
@sabinammm Yes, exactly! Instead of just updating a4, you rebase this whole branch! (and also force-push that) :)

sabinammm:
@fuzzylogic2000 ok, I merged master and tested locally. It works perfectly, except that the links to comments in the export really are still wrong like in the issue https://github.com/liqd/adhocracy-plus/issues/1361
For example, a link to the comment that works is
http://127.0.0.1:8004/liqd-orga/projects/module/umfrage/?comment=7
and the exported link is
http://127.0.0.1:8004/liqd-orga/projects/project-name/?comment=7
which actually links to the project-page.

sabinammm:
@fuzzylogic2000 It works great and the comment links are no longer an issue !!

PR #1365, opened 2021-07-07
jd 2021 07 devtool use source-map in prod

goapunk:
According to the docs `eval` shouldn't be used in production. We use `source-map` in other projects, so this makes it consistent.

### DIFF
--- a/webpack.prod.js
+++ b/webpack.prod.js
@@ -3,7 +3,7 @@ const common = require('./webpack.common.js')
 const TerserPlugin = require('terser-webpack-plugin')
 
 module.exports = merge(common, {
-  devtool: 'eval',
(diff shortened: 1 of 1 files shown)


goapunk:
randomly stumbled upon this and opened this MR so I don't forget about it.

phillimorland:
nice although that fixed here: https://github.com/liqd/adhocracy-plus/pull/1310/files?file-filters%5B%5D=.js but we were waiting to merge it till after an important release not sure what the hold up is now?!

phillimorland:
closed in favour of https://github.com/liqd/adhocracy-plus/pull/1310/files?file-filters%5B%5D=.js

PR #1366, opened 2021-07-08
Jd 2021 07 psycopg pin 2.8

goapunk:


fuzzylogic2000 (approved):
Yeah, this should be fine, but it's not run by the github actions?

goapunk:
@fuzzylogic2000 not until #1358 is merged :)

PR #1373, opened 2021-07-23
ideas/serializers: add user rating

Rineee:
This adds the user rating to idea serializer with rating_id and value. The id is needed to change a users rating, i.e. for put request. If user hasnt rated yet or is not logged in, user rating will be None.

### DIFF
--- a/apps/ideas/serializers.py
+++ b/apps/ideas/serializers.py
@@ -39,6 +39,7 @@ class IdeaSerializer(serializers.ModelSerializer):
     negative_rating_count = serializers.SerializerMethodField()
     labels = LabelListingField(many=True)
     category = serializers.StringRelatedField()
+    content_type = serializers.SerializerMethodField()
     user_rating = serializers.SerializerMethodField()
 
     class Meta:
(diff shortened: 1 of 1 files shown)


khamui:
@Rineee checked out this branch and the corresponding branch in zt:app, but getting this error (404):

Bildschirmfoto 2021-07-26 um 16 20 28

not sure if i miss something, the pk used is from that idea. Do you have an idea what is missing?

Rineee:
> @Rineee checked out this branch and the corresponding branch in zt:app, but getting this error (404):
> 
> Bildschirmfoto 2021-07-26 um 16 20 28
> 
> not sure if i miss something, the pk used is from that idea. Do you have an idea what is missing?

@kha Hm, no I cant see anything wrong at the first glance.. but the idea is not yet rated from the user, right? Because if it is, you would need patch/put, but anyway that should give you 'method not allowed' or 'bad request' instead of 'not found'. 

Can you check what the server says when you rate the idea via Aplus web app? For me, it says for example 
`[02/Aug/2021 09:27:44] "POST /api/contenttypes/82/objects/94/ratings/ HTTP/1.1" 201 210`

I didnt get to installing the app again yet, but as soon as I did, I will check it out!

Rineee:
@kha I now added the content_type to the idea serializer. Will now adjust the api calls in the app PR https://github.com/liqd/zt-app/pull/62

khamui:
LGTM! content_type is there

PR #1385, opened 2021-08-09
Kl 2021 08 check permissions in app serializer

fuzzylogic2000:


fuzzylogic2000:
Just a second, a few more lines to the tests and the missing lines are also covered!

apps/ideas/serializers.py
```diff
@@ -86,6 +91,30 @@ def get_user_rating(self, idea):
                     return RatingSerializer(ratings.first()).data
         return None
 
+    def get_has_rating_permission(self, idea):
+        user = None
+        request = self.context.get('request')
+        if request and hasattr(request, 'user'):
+            user = request.user
+            return user.has_perm('a4_candy_ideas.rate_idea', idea)
+        return False
```
fuzzylogic2000, line 100: > Just a second, a few more lines to the tests and the missing lines are also covered!

Well, even when we have no user logged in, the API gets the anonymous user, so I can't test these lines. But I would still like to keep them as is. So, test coverage is falling. :woman_shrugging:

fuzzylogic2000:
Alright, done! And can be checked together with https://github.com/liqd/zt-app/pull/85
@Rineee @phillimorland @goapunk

apps/ideas/serializers.py
```diff
@@ -86,6 +91,30 @@ def get_user_rating(self, idea):
                     return RatingSerializer(ratings.first()).data
         return None
 
+    def get_has_rating_permission(self, idea):
+        user = None
```
Rineee, line 95: I think this is not needed, right?

fuzzylogic2000: Yes, you're right! Will remove it.

PR #1392, opened 2021-08-10
apps/ideas: enable idea deletion via API

fuzzylogic2000:
to test: get or generate token via
`curl -X POST http://localhost:8004/api/login/ -H 'Accept: application/json;' -d 'username=$username&password=$password'`
to delete idea
`curl -X DELETE http://localhost:8004/api/modules/$moduleId/ideas/$ideaPk' -H 'Accept: application/json;' -H 'Authorization: Token $token'`

### DIFF
--- a/apps/ideas/api.py
+++ b/apps/ideas/api.py
@@ -10,6 +10,7 @@ from .serializers import IdeaSerializer
 
 class IdeaViewSet(ModuleMixin,
                   mixins.CreateModelMixin,
+                  mixins.DestroyModelMixin,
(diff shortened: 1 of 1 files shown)


fuzzylogic2000:
Oh, right, I need to check if I can also do that as user and not only as admin. Sorry, forgot!

fuzzylogic2000:
> Oh, right, I need to check if I can also do that as user and not only as admin. Sorry, forgot!

I couldn't, but that is fixed now!

Rineee (approved):
Yes!!

PR #1394, opened 2021-08-11
ideas/serializers: use right category key from request context in create

Rineee:
needed for https://github.com/liqd/zt-app/pull/88


fuzzylogic2000 (approved):
Cool!

PR #1395, opened 2021-08-16
apps/ideas: add deletion permission check to serilaizer

fuzzylogic2000:


khamui:
LGTM! working with Frontend part!

PR #1422, opened 2021-09-07
apps/module dashboards: no export in nav before module is added to pr…

### LINKED ISSUE #1412: dashboard: export nav is shown when project is published, but module isn't
**URL:** 
**user:** 
**expected behaviour:** The export is only shown when project and module are published.
**behaviour:** For every module the export nav is shown as soon as the project is published. That makes sense when a module is published (added to the project), but is a bit weird when it isn't.
**important screensize:**
**device & browser:** 
**Comment/Question:** like https://github.com/liqd/a4-meinberlin/issues/3837 

Screenshot?
Screenshot from 2021-09-07 10-38-39

sabinammm:
…oject

Closes #1412


fuzzylogic2000 (approved):
Supercool!

PR #1430, opened 2021-09-10
apps/exports: use export mixins from a4

sabinammm:
I did something wrong with this pull request: https://github.com/liqd/adhocracy-plus/pull/1360

I will try it again and hopefully not break everything.... @fuzzylogic2000

### DIFF
--- a/apps/budgeting/exports.py
+++ b/apps/budgeting/exports.py
@@ -3,25 +3,24 @@ from django.utils.translation import ugettext as _
 from rules.contrib.views import PermissionRequiredMixin
 
 from adhocracy4.comments.models import Comment
-from adhocracy4.exports import mixins as a4_export_mixins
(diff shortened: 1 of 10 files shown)


sabinammm:
@fuzzylogic2000 I don't understand this: I tested this locally by merging the current master into the branch. It worked fine. What am I not understanding? The test that worked locally fails during build here.

fuzzylogic2000:
@sabinammm What is the state here? You did already merge the PR, right? So we only need to fix the test?

I guess the test is failing, because we changed the format of the created for the UserGeneratedContentMixin after I did this. But when you rebased this on master and run `make test`, it should have failed as well? Or it could have been in an a4 version before we did that change.

fuzzylogic2000:
There is a PR fixing the test now!

PR #1432, opened 2021-09-10
Sm 2021 09 a4 phase factory and helpers

sabinammm:


fuzzylogic2000 (commented):
Nice! So much work!
You should also delete the helpers from a+, that are not needed anymore: https://github.com/liqd/adhocracy-plus/blob/master/tests/helpers.py

sabinammm:
Thank you! I did forget to get rid of some of the helpers, so I amended the first commit. But I couldn't delete the helpers.py entirely because the assert_template_response function in it is used in some tests, and I didn't find an a4 equivalent. Is there an equivalent that I missed? Otherwise, I would delete all the other functions in the helpers.py and leave that one as it is.

fuzzylogic2000:
Yes, I think we do not have that in a4, but we also use that helper a lot in mB. So another option would be moving it over to a4 and using it in both projects.

sabinammm:
Then I'd like to add the method to a4, update a4 and then merge this after I've refactored completely if that's ok.

fuzzylogic2000:
Looks good! I merged it in a4, so you can update!

fuzzylogic2000 (approved):
Yeah!

PR #1438, opened 2021-09-17
Kl 2021 09 migrate polls

fuzzylogic2000:
Analogous to https://github.com/liqd/a4-meinberlin/pull/3647, but going further as we can use all poll code from a4 directly.

Styling is still missing!


fuzzylogic2000:
Right, this also breaks the comments styling, because we use the template from a4. Instead of overwriting that in here using the new comments, we should overwrite in mB with the old ones?! And change the template in a4. Are we also overwriting the poll detail in opin?

fuzzylogic2000:
@phillimorland @Rineee @sabinammm @khamui Could you all test it? And then if the DBs are fine and the same question, votes and comments are shown as before, we can merge and style afterwards?

phillimorland:
@fuzzylogic2000 mine is working fine but I had an almost empty database, so wam hesitent to merge, i'll work on here for styling and if someone with an older database can also test that would be better

sabinammm:
I tested it also with a database with a few things and some comments. It works fine for me. I still don't really understand this, but I did notice that the "content_type": had a different value after migration.

fuzzylogic2000:
So, let's wait for at least @Rineee to also test. :)

Rineee (commented):
Yes, its looking good and works well!

Issue #1437, opened 2021-09-16
ap poll: Saved poll still not finished after saving


fredewe:
**URL:** https://aplus-dev.liqd.net/frede-test/dashboard/modules/umfrage-19/poll/
**user:** Initiator
**expected behaviour:** When i save my poll i expect the red sign in the sidebar to disappear
**behaviour:** When i save my poll i have to manually reload the page to see that i filled out everything
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?
![Bildschirmfoto 2021-09-16 um 11 19 03](https://user-images.githubusercontent.com/5470479/133586739-9a9da4d5-0523-4ed6-842d-48fafb85119e.png)

lthueer:
@phillimorland this issue occurs even with the new poll module:
![Bildschirmfoto 2021-09-22 um 16 29 48](https://user-images.githubusercontent.com/36957474/134363113-8bb52af0-1e86-467d-910a-02622f985675.png)

phillimorland:
Damm! will check it out

PR #1453, opened 2021-09-23
dashboard/editpolls/save: adding selector to nav to update on success

khamui:
closing #1437


fuzzylogic2000 (commented):
For me locally this still doesn't work.

khamui:
@fuzzylogic2000 I just doublechecked (with not linked a4) and it works for me 🤔btw it is a bit delayed, it does not happen immediately, though.

fuzzylogic2000:
Yes! You are right! Today it worked fine. :blush:

PR #1454, opened 2021-09-23
Ks 2021 09 add poll export

Rineee:
This only works with liqd/adhocracy4#845

Do not merge until liqd/adhocracy4#845 is merged and A4 dependency updated here.


fuzzylogic2000 (approved):
Same here: more yeah!

PR #1458, opened 2021-09-27
package.json: dependencies rm react markdown packages as loaded in a4

phillimorland:
a4 must be updated after this has been merged https://github.com/liqd/adhocracy4/pull/850


khamui:
> @goapunk @khamui @fuzzylogic2000 is there a reason we shouldn't do this? I feel it is better to install packages in a4 that we use in a4 but i still feel weird about it?!

i agree, think that makes sense and is better! I can't see any reason why we shouldn't do this, anyway let's see what the others say :)

fuzzylogic2000:
I also think it's better to add them in a4 and not in the projects if that's possible. (I thought the reason we installed it everywhere was because it wasn't.) And another reason I could think of would be if we don't use it on all platforms, but when it comes with a4 we probably do.

PR #1464, opened 2021-09-28
contact/form/checkbox: removing style for form-check

khamui:
@phillimorland i checked if this removal of `.form-check` breaks for instance on the `signup page`, but it seems that because there the `<input>` is nested inside of the `<label>` it doesn't receive the breaking `widht: 100%` anyway.

on the contact form it is not nested, but next to each other...

```html
<div class="form-check">
    <Input />
    <label></label>
</div>
```


phillimorland:
not sure i'm getting you, so as it stands this pull request doesn't reduce the width, is that right?

PR #1466, opened 2021-09-28
style.scss: update config to only load required components, add fix t…

phillimorland:
…o variables


phillimorland:
some small styling issues to iron out but it is mostly working! :D

PR #1471, opened 2021-09-29
component library: remove translation tags

fuzzylogic2000:
fixes https://sentry.liqd.net/organizations/liqd/issues/1624/?project=30


khamui:
makes sense!

PR #1472, opened 2021-09-29
dashboard/nav: store expand states in cookie

khamui:


adhocracy-plus/assets/js/app.js
```diff
@@ -8,6 +8,7 @@ import '../../../apps/dashboard/assets/ajax_modal.js'
 import '../../../apps/maps/assets/map-address.js'
 import '../../../apps/moderatorremark/assets/idea_remarks.js'
 import '../../../apps/newsletters/assets/dynamic_fields.js'
+import { initDashboardAccordeon } from '../../../apps/dashboard/helpers'
```
phillimorland, line 11: Accordeon is spelt Accordion, sorry! I think it's also probably wrong in mB but just for code searches it's better to have it correct

adhocracy-plus/assets/js/app.js
```diff
@@ -59,16 +60,14 @@ function init () {
   // This function adds required classes to iframes added by ckeditor
   $('.rich-text iframe').addClass('ck_embed_iframe')
   $('.ck_embed_iframe').parent('div').addClass('ck_embed_iframe__container')
+
+  // Initialize observers for dashboard dropdowns
+  initDashboardAccordeon()
 }
```
phillimorland, line 66: I think it might be better to only load this on dashboard as it's not needed anywhere else, what do you think?

phillimorland (changes requested):
Yo I would just change spelling and let me know what you think about the load. 
Also I just noticed that there is this error in console 
`Uncaught Error: Syntax error, unrecognized expression: ##dashboardCollapseProj`
do you get that?

adhocracy-plus/assets/js/init_dashboard_accordion.js
```diff
@@ -1,6 +1,9 @@
 import Cookies from 'js-cookie'
 
-export function initDashboardAccordion () {
+document.addEventListener('DOMContentLoaded', initDashboardAccordion, false)
```
khamui, line 3: this is added because otherwise it will fire the function too early.

phillimorland:
> Yo I would just change spelling and let me know what you think about the load. Also I just noticed that there is this error in console `Uncaught Error: Syntax error, unrecognized expression: ##dashboardCollapseProj` do you get that? works beautifully though! :)

still getting that error in console? i assume it's the double hash but notsure where/why? it's working so also fine to merge but would be nice to not have additional console errros for release as we already have too many :( @khamui

phillimorland (commented):
if console error isn't easy fix then merge as it's all there and is great!

khamui:
> > Yo I would just change spelling and let me know what you think about the load. Also I just noticed that there is this error in console `Uncaught Error: Syntax error, unrecognized expression: ##dashboardCollapseProj` do you get that? works beautifully though! :)
> 
> still getting that error in console? i assume it's the double hash but notsure where/why? it's working so also fine to merge but would be nice to not have additional console errros for release as we already have too many :( @khamui

ah sorry i didn't reply to this. but I tested on my side and i don't get this error. Anything special you did to trigger it?

PR #1486, opened 2021-10-05
locale-source: translation strings as pushed to transifex

Rineee:


sabinammm (commented):
Do I understand this correctly that these files are deleted because they're essentially parked somewhere else? Is it relevant that some translations seem still to be missing or is that just part of another step?
image

fuzzylogic2000 (approved):
Yeah!

fuzzylogic2000:
> Do I understand this correctly that these files are deleted because they're essentially parked somewhere else? 

The files aren't deleted, but changed. That's like in the other projects.

> Is it relevant that some translations seem still to be missing or is that just part of another step? 

And that's what's different from mB for example. Here we use transifex, so we do not do the translations, but push and pull them.
So, the steps are:
- make new po files with the new strings 
- push these to transifex
- on transifex other people translate
- then we pull the strings and make a PR with the translations.
What @Rineee did was the first two steps and to have the same strings in the code as on transifex, she opened this PR. (and before she made the new po files she also pulled the strings from transifex to make sure, we do not overwrite any translations when pushing. So the newly added strings are from this step.)
@sabinammm

PR #1519, opened 2021-10-19
style.scss: update config to only load required components, add fix t…

### LINKED ISSUE #1622: debate list: list tile is clicked when dropdown ordeing filter clicked
**URL:** https://aplus-dev.liqd.net/liqd-tests/projects/test-the-debate/
**user:** 
**expected behaviour:** 
**behaviour:** When I try to order by "most comments", instead of choosing the filter in the dropdown, the debate subject tile is clicked and I am taken to the detail page.
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?

fuzzylogic2000:
…o variables

check out https://github.com/liqd/adhocracy-plus/pull/1466

fixes #1622


fuzzylogic2000 (commented):
@phillimorland Is this still a thing that we want to do? Then maybe now with the updates and the extensive testing is a good idea? (You originally opened the PR, but now it is in my name because of the master/main rename)

phillimorland:
@fuzzylogic2000 I was wondering where that was :p yes I would still like to try it as it's possible bootstrap causes the map upload thing, will try to get it up to date this morning!

fuzzylogic2000 (commented):
Looks good, also in the browser. 
I cannot reproduce the failing test locally. :(

fuzzylogic2000:
> Looks good, also in the browser. I cannot reproduce the failing test locally. :(

But the test is not failing anymore after running them again. :)

PR #1521, opened 2021-10-19
WIP: Kl 2021 05 pm block clean

fuzzylogic2000:
check out https://github.com/liqd/adhocracy-plus/pull/1295


phillimorland:
@fuzzylogic2000 is the issue here the migration from editing blocks which are used? as we will look at wagtail anyway should we maybe try and take the easy parts of this?

fuzzylogic2000:
@phillimorland Well, I think, we have to start again anyway, but let's keep this open to have a look and close afterwards!

PR #1523, opened 2021-10-19
Ks 2021 04 remove old questions and likes apps

fuzzylogic2000:
check out https://github.com/liqd/adhocracy-plus/pull/1242

before merge, check if questions and likes tables are dropped in

- [x] kosmo
- [x] diid -> after rebase


fuzzylogic2000:
And when at it, also drop the polls tables! They were emptied with #1639

phillimorland:
@fuzzylogic2000 how is this going? we already did a diid rebase right?

fuzzylogic2000:
@phillimorland Yes, this is all fine! I will just do a rebase of this PR and it should be good to go!

fuzzylogic2000:
@Rineee I had to remove a dependency from a migration. As these should all be run in all projects already and also it shouldn't be a problem on a fresh DB, it should be fine. Could you also check on your local DB, though?

fuzzylogic2000:
@Rineee I had to remove a dependency from a migration. As these should all be run in all projects already and also it shouldn't be a problem on a fresh DB, it should be fine. Could you also check on your local DB, though?

Rineee:
> @Rineee I had to remove a dependency from a migration. As these should all be run in all projects already and also it shouldn't be a problem on a fresh DB, it should be fine. Could you also check on your local DB, though?

@fuzzylogic2000 Yes, it works all fine with empty and non-empty DB! I will fix the tests!

Rineee:
> > @Rineee I had to remove a dependency from a migration. As these should all be run in all projects already and also it shouldn't be a problem on a fresh DB, it should be fine. Could you also check on your local DB, though?
> 
> @fuzzylogic2000 Yes, it works all fine with empty and non-empty DB! I will fix the tests!

@fuzzylogic2000 Ok, tests are fixed! But now I am confused why I left the empty likes/ and questions/ folders (or even the models.py in questions/)?? I dont remember what I did :exploding_head: :sweat_smile: Do you think we should just also remove it? Or we do it next time.. but apart from my confusion, this is good to go!

fuzzylogic2000:
> > > @Rineee I had to remove a dependency from a migration. As these should all be run in all projects already and also it shouldn't be a problem on a fresh DB, it should be fine. Could you also check on your local DB, though?
> > 
> > 
> > @fuzzylogic2000 Yes, it works all fine with empty and non-empty DB! I will fix the tests!
> 
> @fuzzylogic2000 Ok, tests are fixed! But now I am confused why I left the empty likes/ and questions/ folders (or even the models.py in questions/)?? I dont remember what I did exploding_head sweat_smile Do you think we should just also remove it? Or we do it next time.. but apart from my confusion, this is good to go!

@Rineee Yes, I think we should also remove that! Thought, we already did? Maybe I fucked up the rebase.

Rineee:
> And when at it, also drop the polls tables! They were emptied with #1639

Yes, I will do that!

Rineee:
> > And when at it, also drop the polls tables! They were emptied with #1639
> 
> Yes, I will do that!

@fuzzylogic2000 Ok, I now only deleted the old poll migrations, but left the poll app as we still need the export view and dashboard components to be overwritten.

fuzzylogic2000 (commented):
Yes, it works fine and all is still as before!

PR #1545, opened 2021-10-28
livequestions/affiliation/dropdown: transfering fix from kosmo and ma…

khamui:
…ke 'all' selection working correctly

this fixes liqd/a4-kosmo#92 in aplus


phillimorland (approved):
Nice!

PR #1575, opened 2021-11-15
apps/cms: add helptexts for site-wide organisation pages settings

fuzzylogic2000:
I did add this in kosmo... But I think, it makes more sense to (also) have it here!


phillimorland (approved):
LGTM :)

PR #1586, opened 2021-11-19
Sm 2021 11 update makemessages

### LINKED ISSUE #1585: Update makemessages command in Makefile
**URL:** 
**user:** 
**expected behaviour:** When adding new texts for translation, `make po` in the makefile uses the `makemessages` command to pull new texts into the localization file.
**behaviour:** The command no longer works and throws a CommandError. 
**important screensize:**
**device & browser:** 
**Comment/Question:** 
The issue was fixed [here](https://github.com/liqd/a4-meinberlin/pull/3974#issuecomment-974015445) already.
Screenshot?

sabinammm:
Fixes #1585

The application (**2nd commit**) of the changed `make po` script (**1st commit**) led to some duplicate lines in the po files. 

These were left untouched because they are likely a result of the changed makemessages command, and would thus be generated anytime make po is executed in future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,8 +123,8 @@ lint-fix:
 
 .PHONY: po
 po:
-	$(VIRTUAL_ENV)/bin/python manage.py makemessages -d django --extension html,email,py --ignore '$(CURDIR)/node_modules/adhocracy4/adhocracy4/*'
-	$(VIRTUAL_ENV)/bin/python manage.py makemessages -d djangojs --ignore '$(VIRTUAL_ENV)/*' --ignore '$(CURDIR)/node_modules/dsgvo-video-embed/dist/*'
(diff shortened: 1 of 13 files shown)


fuzzylogic2000 (approved):
Looks good!
I think @Rineee fixed the double translation lines somewhere? Maybe we can also do that here?

And we should get transifex into the shame shape as a+. So. we should probably pull strings before we do the `makemessages`.

Rineee:
> Looks good! I think @Rineee fixed the double translation lines somewhere? Maybe we can also do that here?
> 
> And we should get transifex into the shame shape as a+. So. we should probably pull strings before we do the `makemessages`.

Yeah, I fixed it here, its the --ignore, but that seems not to work anymore with the --all?

PR #1587, opened 2021-11-22
Sm 2021 11 fix make translations

### LINKED ISSUE #1585: Update makemessages command in Makefile

sabinammm:
Fixes #1585

The makefile is

1. fixed in the first commit 
2. applied in the second commit with a pull from transifex. This was then also pushed to transifex.

as discusssed [here](https://github.com/liqd/adhocracy-plus/pull/1586#pullrequestreview-812165389) .

A safety branch of the status of transifex was created beforehand [here](https://github.com/liqd/adhocracy-plus/tree/sm-2021-11-transifex-safety).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,8 +123,8 @@ lint-fix:
 
 .PHONY: po
 po:
-	$(VIRTUAL_ENV)/bin/python manage.py makemessages -d django --extension html,email,py --ignore '$(CURDIR)/node_modules/adhocracy4/adhocracy4/*'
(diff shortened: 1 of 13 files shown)


Rineee (approved):
Nice! I closed the other pull request, so we dont get confused..

PR #1625, opened 2022-01-07
apps/templates: added general error message to top of create and upda…

### LINKED ISSUE #1597: Missing error messages at the top of the page for missing description in create & edit idea forms
**URL:** https://aplus-dev.liqd.net/sabina-test/mapideas/create/module/brainstorming-mit-karte-15/
**user:** any user who creates or edits and idea and leaves the description field empty or empties it when editing.
**expected behaviour:** I get an error message on the top of the page calling attention to an error actually occurring lower on the page
**behaviour:** There is only an error message at the bottom of the page.
**important screensize:** any
**device & browser:** ubuntu chrome
**Comment/Question:** @fredewe  should this be the same as in other forms with an additional error message at the top?

Screenshot?
**How it is:**
image

**Should it be like this?**
image

sabinammm:
…te forms

Fixes #1597

This was already done in meinBerlin [here](https://github.com/liqd/a4-meinberlin/pull/3985/files).

### DIFF
--- a/apps/budgeting/templates/a4_candy_budgeting/proposal_create_form.html
+++ b/apps/budgeting/templates/a4_candy_budgeting/proposal_create_form.html
@@ -23,6 +23,9 @@
                 </nav>
 
                 <h1>{% trans 'Submit a new proposal for this project'%}</h1>
+                {% if form.errors %}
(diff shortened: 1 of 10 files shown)


fuzzylogic2000 (approved):
very nice!

PR #1628, opened 2022-01-10
Pm 2022 01 map issues

phillimorland:


apps/maps/assets/map_choose_polygon_with_preset.js
```diff
@@ -166,7 +168,7 @@ function init () {
     },
 
     _addToMap: function (map, shape) {
-      $('#map-import-modal').modal('hide')
+      $('#map-import-modal').hide()
       // work around boostrap backdrop bug
```
phillimorland, line 172: this is a hack, i was hoping to find a fix that would work with a4 version but didn't and too much time has been spent, will add issue to investigate further when there is more time

fuzzylogic2000 (approved):
Looks good!

PR #1631, opened 2022-01-10
apps/rules: use new permission from a4 and add tests

fuzzylogic2000:


fuzzylogic2000:
Alright, now it is done, coverage is not falling anymore. @Rineee Could you have a look?

Rineee (approved):
Very nice and much clearer!

PR #1639, opened 2022-01-11
apps/polls: remove old polls, questions, choices and votes from DB, d…

fuzzylogic2000:
…elete models

The poll models and tables were migrated to a4 in the last release.
This deletes the models and tables that are not used anymore.


phillimorland:
installs and works for me but don't have any old polls to check it on as had to scrap my database a while back, @Rineee can you test it?

phillimorland (approved):

Rineee (approved):
Yeah, supercool! I even checked with postgres db and all is fine and the tables gone :tada: 

Should we then also do that for the likes tables?

PR #1623, opened 2022-01-06
settings: add variables for ck editor embed provider

fuzzylogic2000:
fixes #1546 
But let's see if we need it or if the server setting changes are sufficient.

PR #1643, opened 2022-01-12
Revert "settings: add variables for ck editor embed provider"

fuzzylogic2000:
Reverts liqd/adhocracy-plus#1623


fuzzylogic2000:
@phillimorland @goapunk You were right, it is still working without this commit, so I will merge this. Should we also move the settings somewhere else like to the dev-settings to make clear, that the prod settings come from somewhere else?

PR #1699, opened 2022-02-03
docs/ckeditor: add docs for ckeditor

phillimorland:


phillimorland:
I don't write much documentation and i'm better with examples then instructions so let me know if it's too vague and i'll try again

docs/ckeditor.md
```diff
@@ -0,0 +1,24 @@
+# CKEditor
+
+We use the ckeditor in the dashboard in order for users to add formatted text, editable images, accordions and videos. To achieve this we have 4 configured editors for different use cases, default, image-editor, collapsible-image-editor, video-editor.
```
fuzzylogic2000, line 3: Do we currently also use the default one without images for ideas?

fuzzylogic2000 (approved):
Nice! I think it is a perfect start and we can add to it whenever we stumble over more stuff that is confusing or needs explanation.

docs/ckeditor.md
phillimorland: yep, we use image upload widget for images in ideas

fuzzylogic2000 (approved):
Nice!

PR #1705, opened 2022-02-08
assets/app.js: do not display the image input field for adding via url

phillimorland:


phillimorland:
@fuzzylogic2000 we wanted quick and dirty :( i will try one last time withe the api but I think for now that this is fine, I did a check on the link dialog box and it was all good but the guys will need to test thouroughly

fuzzylogic2000 (commented):
Now my whole image uploading is not working anymore, no matter if I select an old one or upload a new one, it will not show it. :cry:

phillimorland:
@fuzzylogic2000 so the url input is required for the database path, not just a url so actually only way to do it is to hide it so this should be fine, I tested and it does actually add image this time :p

fuzzylogic2000 (approved):
Hooray! It's gone and works! :)

PR #1710, opened 2022-02-11
ideas/api: delete image of idea on update

Rineee:


khamui:
nice! removing image works!

PR #1739, opened 2022-03-03
comments_async/date: break date into new line

khamui:


Rineee:
@khamui Oh, did we forget about this? This fixes #1787, right?

khamui:
@Rineee oh, yes oops..

phillimorland:
doesn't this need a4 update before it's ready for review?

khamui:
oh okay, should have removed it back then. so generally we remove this label, when a4 (that this is depending on) was merged right?

phillimorland:
@khamui for a4 depending pull requests it's a bit messy as you can't really test one without the other so you should always have a4 label and always include the link to the a4 PR in the depending PR so they can be tested together, if it's fine then the tester can then merge the a4 one and either inform the creator that their app PR can be merged after they update the a4 hash or the tested can do it themselves if they have time, it's not really a perfect system, sorry!

PR #1779, opened 2022-04-07
jest: adding framework, configs, mocks and commands

khamui:
this adds the testing framework jest and configures everything to
run JS tests in adhocracy plus.

And one component test is being added.


fuzzylogic2000 (approved):
Supercool!

PR #1789, opened 2022-04-13
Jd 2022 04 node 16

goapunk:
work towards node 16 compatibility

* downgrade deps to consistently depend on react 17
* add alias for no longer working react-slick and react-popper
* change ci


phillimorland:
@goapunk whats the status here? both packages have recent updates maybe they are now compatible?

goapunk:
@phillimorland I rebased it and removed the no longer needed workarounds, however this is still blocked by a4 updates. I think we need a different update strategy here. We shouldn't do a+ before a4 updates (e.g. a4 still using react 17 and on aplus we already have packages depending on react 18, otherwise it will become really annoying with node 16.

phillimorland:
@goapunk we good now?

PR #1796, opened 2022-04-21
Kl 2022 04 issues

fuzzylogic2000:
Translations for the email can be taken from https://github.com/liqd/a4-meinberlin/blob/main/meinberlin/templates/account/email/unknown_account.de.email


goapunk (approved):
nice :+1:

PR #1797, opened 2022-04-21
tests/notifications: add tests for notifications

fuzzylogic2000:
After merging and rebasing, we can also add the tests for the new notifications in kosmo...


Rineee (approved):
Superdupercool!

Issue #1737, opened 2022-03-01
sending newsletter from empty organisations throws 500


goapunk:
**URL:*https://aplus-dev.liqd.net/julian-newsletter-no-projects/dashboard/newsletters/create/* 
**user:*initiator/admin* 
**expected behaviour:*trying to send a newsletter without selecting a project will show a warning* 
**behaviour:*newsletter can be sent and server returns 500* 
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?

PR #1800, opened 2022-04-22
Kl 2022 04 newsletter project

fuzzylogic2000:
- adds newsletter tests
- second commit is a4-depending (needs to be updated after https://github.com/liqd/adhocracy4/pull/1057 is merged) -> done and fine!

- [x] but what I set of to do, which is fixing this #1737 is still missing

My explanation about why it did not work was all wrong. Bit now it does work and fixes the issue!


goapunk (approved):
very nice!

PR #1801, opened 2022-04-25
apps/contrib/assets: removing Alert and ErrorList (using from a4)

khamui:


fuzzylogic2000 (approved):
Whoop whoop!

khamui:
not sure if there is a more elegant way to set that `--passWithNoTests` flag to the coverage action. Maybe @goapunk knows it?  Generally i am not sure if doing all the js/frontend testing if there are no tests makes sense? Also having jest but no tests shouldn't be the usual case...

PR #1809, opened 2022-04-26
tests/follows: move follow factory in a4 - a4 depending

fuzzylogic2000:
depends on https://github.com/liqd/adhocracy4/pull/1065


goapunk (approved):
looks good, do we do the a4 update here or in a separate PR?

fuzzylogic2000:
> looks good, do we do the a4 update here or in a separate PR?

@goapunk Here!

And I will also add a name change for a weirdly named test. :)

goapunk (approved):
cool!

PR #1812, opened 2022-04-26
assets/* templates/* : update list items in multiple templates to fix…

phillimorland:
… design issues, update component lib, update btn for better a11y


khamui:
nice! looking good. verified in component library

PR #1823, opened 2022-05-03
requirements: use different psycopg2 flavors

goapunk:
* use psycopg2 for prod
* use psycopg2-binary for dev


goapunk:
according to the psycopg2 docs `psycopg2-binary` should not be used in production

fuzzylogic2000 (approved):
Nice!

PR #1833, opened 2022-05-16
[5991 + 5987] add model and form mixin for users's agreement on orgsnisation's terms of use

fuzzylogic2000:
I put it all into the mixin instead of using rules.


apps/organisations/mixins.py
```diff
@@ -0,0 +1,48 @@
+from django import forms
+from django.utils.translation import gettext_lazy as _
+
+from apps.organisations.models import OrganisationTermsOfUse
+
+ORGANISATION_TERMS_OF_USE_LABEL = _(
+    'Yes, I have read and agree to the terms of use. '
```
fuzzylogic2000, line 7: The label probably has to go to the __init__, so that we can actually add the link to the organisation's terms of use.

apps/contrib/templates/a4_candy_contrib/includes/form_checkbox_field.html
```diff
@@ -3,7 +3,7 @@
 <div class="form-check">
     <label class="form-check__label">
         {{ field }}
-        {{ field.label }}
+        {{ field.label }}{% if field.field.required %}<span role="presentation" title="{% trans 'This field is required' %}">*</span>{% endif %}
```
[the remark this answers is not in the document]
phillimorland: no clue but only place I know we have a required checkbox, (sign up form) we don't use it?!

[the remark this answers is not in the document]
phillimorland: I think it's from mB so we could check there?

phillimorland (commented):
Code looks good but will wait for Rine for proper code review.
When testing and attempt to save without checking box it only has the browser error handling, is this intended? also maybe this is why we didn't have required on the checkbox as when we have error handling this would mean there would double errors?

fuzzylogic2000:
@phillimorland I changed the error handling and added the link to the label. Adding the second sentence as a helptext (not quite the design) is confirmed by @mcastro-lqd .
Sooo, we could merge to try it out. But then this would get half-finished into kosmo. :(

phillimorland (approved):
fabulous error handling :)

apps/organisations/mixins.py
```diff
@@ -0,0 +1,76 @@
...
+                )
+                agreed.has_agreed = \
+                    self.cleaned_data['organisation_terms_of_use']
+                agreed.save()
+            except OrganisationTermsOfUse.DoesNotExist:
+                OrganisationTermsOfUse.objects.create(
+                    user=self.user,
+                    organisation=self.module.project.organisation,
+                    has_agreed=self.cleaned_data['organisation_terms_of_use']
+                )
```
Rineee, line 60: Maybe this could be handled with get_or_create()?

apps/organisations/mixins.py
```diff
@@ -0,0 +1,76 @@
...
+        self.fields['organisation_terms_of_use'].label = \
+            self._get_terms_of_use_label
+
+        if OrganisationTermsOfUse.objects.filter(
+            user=self.user,
+            organisation=self.module.project.organisation
+        ).exists() and OrganisationTermsOfUse.objects.get(
+            user=self.user,
+            organisation=self.module.project.organisation
+        ).has_agreed is True:
```
Rineee, line 30: Wouldn't it be the same to just check if the entry with has_agreed=True exists? Then it would only be one query.. So I mean
`if OrganisationTermsOfUse.objects.filter(
            user=self.user,
            organisation=self.module.project.organisation,
            has_agreed=True
        ).exists()`

Rineee: And maybe it could go into the model as a property or method? Sth like user_has_agreed(user, organisation)? But not sure if that makes sense.. just thought we will probably need this again at other places.

Rineee:
@fuzzylogic2000 Supercool, it works great! Just made some proposals code wise, but I think this is the way to go!

apps/contrib/templates/a4_candy_contrib/includes/form_checkbox_field.html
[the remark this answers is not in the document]
Rineee: @fuzzylogic2000 Was it maybe, because for the image_right_of_use, it is only required if there is an actual image (and that is checked in the clean)? So maybe for organisations_terms_of_use the form field could be required?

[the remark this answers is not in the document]
Rineee: Haha, ok, maybe should have read this conversation til the end, sorryyy :see_no_evil:

apps/ideas/views.py
```diff
@@ -109,6 +109,11 @@ class IdeaCreateView(AbstractIdeaCreateView):
     permission_required = 'a4_candy_ideas.add_idea'
     template_name = 'a4_candy_ideas/idea_create_form.html'
 
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        kwargs['user'] = self.request.user
+        return kwargs
+
```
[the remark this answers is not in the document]
fuzzylogic2000: Yes, it was there first, but then all the tests are failing because the mixin is not yet there. We can either add it to all forms right away or keep it like that, test in idea on dev and then move it back when we add it to all the other modules.

apps/organisations/mixins.py
fuzzylogic2000: Yes, both good ideas! Which one should I do now? :)
Will start with trying the second version.

apps/organisations/mixins.py
Rineee: Ah ok, didn't know that! I thought that this is exactly what get_or_create() does, so try: get, except: create. But if this is more robust, lets leave it like that.

apps/ideas/views.py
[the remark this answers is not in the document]
Rineee: Sure, we can also move it there later, didn't think about failing tests!

apps/organisations/mixins.py
Rineee: Should we then also remove get_or_create in other places, you think?

apps/organisations/mixins.py
fuzzylogic2000: I put it into the user model, because that is where we also should put the list for the overview.

Rineee: Perfect!

apps/users/models.py
```diff
@@ -148,3 +149,10 @@ def get_full_name(self):
 
     def get_absolute_url(self):
         return reverse('profile', args=[str(self.username)])
+
+    def has_agreed_on_org_terms(self, organisation):
+        return OrganisationTermsOfUse.objects.filter(
+            user=self,
+            organisation=organisation,
+            has_agreed=True
+        ).exists()
```
Rineee, line 158: Fabulous!

apps/organisations/mixins.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop('user')
+        super().__init__(*args, **kwargs)
+
+        self.fields['organisation_terms_of_use'].label = \
+            self._get_terms_of_use_label
+
+        if self.user.has_agreed_on_org_terms(
+                self.module.project.organisation):
+            del self.fields['organisation_terms_of_use']
```
Rineee, line 26: So clever to just delete the field then!

Rineee (approved):
Yeah, super nice!

PR #1848, opened 2022-05-19
desp: planifolia remove unused css classes

khamui:
the branch with a lot of file changes i can't push, because it does not pass husky :(
but i have it locally, in case we want to have a look i can create a patch i guess.

current state: need replacement mixins/functions for commented out:
`grid-*`
`contrast()`
`mix()`
`luma()`

### DIFF
--- a/adhocracy-plus/assets/scss/_form.scss
+++ b/adhocracy-plus/assets/scss/_form.scss
@@ -21,7 +21,7 @@ input {
     max-width: 100%;
     background-color: $body-bg;
     color: contrast-color($body-bg);
-    box-shadow: 0 0 1px $brand-primary-tint inset;
+    // box-shadow: 0 0 1px $brand-primary-tint inset;
     border-radius: 0;
     border-width: 1px;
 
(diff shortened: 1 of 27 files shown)


phillimorland:
@khamui just to show you what I meant when we talked, instead of deleting references to code we need like you did in the other pr, when updating or removing packages like this, I find it best to just comment out everything that is causing failed install. Once its installed then go through it function by function or whatever and fix what has been commented out for like for like code/functions. We can talk about this again as we need to deal with it during next sprint.

khamui:
oh, this PR was rather a working (with planifolia) but removed the definitely unused classes. Because i didn't know how soon we will make the bigger shift.

but yeah, its not a lot that is unused.

fuzzylogic2000:
@phillimorland @khamui What is the status? Should we merge before release or rather not?

phillimorland:
@fuzzylogic2000 no this will be touched again when we do the button update story https://taiga.liqd.net/project/liqd-liquid-product/us/254?no-milestone=1

PR #1852, opened 2022-05-23
add peer dependencies for a4

goapunk:
depends on https://github.com/liqd/adhocracy4/pull/1044 (needs a4 update)


fuzzylogic2000 (commented):
It did work locally! But obviously installing does not work here. Don't know what to do about that.

fuzzylogic2000:
> @fuzzylogic2000 for some reason we double install react-popper in a4 which may lead to issues, I can have a look at it if you like?

@phillimorland That would be awesome! Thank you!

PR #1854, opened 2022-05-23
deps: downgrade to recommended version

khamui:


package.json
```diff
@@ -49,7 +49,7 @@
     "@babel/preset-env": "7.17.12",
     "@babel/preset-react": "7.17.12",
     "@babel/runtime": "7.17.9",
-    "@testing-library/react": "13.2.0",
+    "@testing-library/react": "12.1.2",
```
goapunk, line 52: latest version is `12.1.5` afaik, Is there a reason why you picked `12.1.2` ?

khamui: i think latest is 13.2.0, but i guess @fuzzylogic2000 used `12.1.2` because it is the one which works with `@testing-library/jest-dom` (in a4-kosmo)?

khamui: e.g. speaking of this:
https://github.com/liqd/a4-kosmo/pull/246/commits/a43e11499a60f21f4cc9d21997b5dbdcd7b665ea

i tested 12.1.5 and this should working fine as well, will add to a4-kosmo and change this

goapunk:
cool, thanks!

package.json
fuzzylogic2000: Never think what I do makes sense. I only added that version because it was Friday and I wanted to leave. :)

PR #1864, opened 2022-05-24
WIP Pm 2022 05 captcheck fix

phillimorland:


adhocracy-plus/templates/account/signup.html
```diff
@@ -53,6 +53,7 @@ <h1>{% trans "Register" %}</h1>
         {% endif %}
 
         {% include 'a4_candy_contrib/includes/form_field.html' with field=form.captcha %}
+        {{ settings.a4_candy_cms_settings.OrganisationSettings.contacts | richtext }}
```
phillimorland, line 56: @fuzzylogic2000 This is not ideal but other uses of settings links in help-text were all done in the html, I tried getting it into the modal help text but not very hard as I'm not sure how to get the correct string out of the rich-text to suit the sentence anyway so I though this was the best bet but pretty sure i'm wrong, or we make a setting just for support?
Maybe I can use `bleach` and only allow the link? that will only work if email is always added as a `mailto` link which probably isn't a good idea?! err brain is foggy sorry!

apps/cms/contacts/models.py
```diff
@@ -35,7 +35,9 @@ def formfields(self):
         # Add captcha to formfields property
         fields = super().formfields
         fields['captcha'] = CaptcheckCaptchaField(
-            label=_('I am not a robot'))
+            label=_('I am not a robot'),
+            help_text=_('If yu are having difficulty please contact us, '
```
goapunk, line 39: typo: yu -> you

apps/cms/contacts/models.py
```diff
@@ -32,15 +34,22 @@ class FormField(AbstractFormField):
...
     @property
     def formfields(self):
+        contacts = self.get_contacts_for_default_site()
         # Add captcha to formfields property
         fields = super().formfields
         fields['captcha'] = CaptcheckCaptchaField(
             label=_('I am not a robot'),
             help_text=_('If you are having difficulty please contact us, '
-                        'details can be found adjacent.'))
+                        'details can be found adjacent.') + contacts)
```
Rineee, line 43: Yeah, so this is the richtext as str.. not sure if that helps?

phillimorland: ooooo, danke i'll give it a go

phillimorland: @Rineee dumb question to add this to the signup form I just change self to form or do I need to restructure the class?

apps/users/forms.py
```diff
@@ -41,7 +41,11 @@ class DefaultSignupForm(SignupForm):
                     'additional information via email.'),
         required=False
     )
-    captcha = CaptcheckCaptchaField(label=_('I am not a robot'))
+    captcha = CaptcheckCaptchaField(
+        label=_('I am not a robot'),
+        help_text=_(
+            'If you are having difficulty please find our contact details '
+            'below.'))
```
phillimorland, line 48: This is where I wanted it, sorry it was form.py not modal

apps/cms/contacts/models.py
phillimorland: NOTE: this solution retains all the html tags, would need to be cleaned first

PR #1868, opened 2022-05-24
Pm 2022 05 deps update

phillimorland:
on surface seems to be simpler then anticipated, install without too many warnings, we will have to be careful about the v2 update but this seems good to go


fuzzylogic2000 (approved):
Nice! Works!

PR #1869, opened 2022-05-25
modules/tiles: using combination of overflow-wrap hyphens and smaller…

### LINKED ISSUE #1780: long title on tile for module: line break within word
**URL:** https://aplus-stage.liqd.net/base/projects/online-konsultation-geschaftsordnung/
**user:** initiator filling in verbal information
**expected behaviour:** Everything that I write looks nice when handed out, no line breaks within words.
**behaviour:** Long, but used word like Geschäftsordnung have a line break within the word
**important screensize:** Desktop
**device & browser:** Mac firefox, chrome
**Comment/Question:** 

Screenshot?

grafik

khamui:
… fontsize to asssure proper word wrapping.

This is one approach to close #1780 

As spoken with @mcastro-lqd the font-size used is too big. So i am using `<h4>` rather than modifying `<h3>` (there is a systematic behind it with screen sizes etc.)

Then to word-wrap the css property `overflow-wrap` is used to not allow wrapping on weird places.

Additionally using css property `hyphens`, but only for smaller screen sizes, otherwise it will wrap quite earlier and look weird. And also languages are not very well supported in browsers other than english/german (see caniuse.com)

### DIFF
--- a/adhocracy-plus/assets/scss/components/_participation-tile.scss
+++ b/adhocracy-plus/assets/scss/components/_participation-tile.scss
@@ -39,6 +39,12 @@
 
 .participation-tile__content {
     padding: 1.5 * $padding;
+    overflow-wrap: normal;
+    hyphens: auto;
+
+    @media (min-width: $breakpoint-lg) {
+        hyphens: none;
+    }
 }
 
 .participation-tile__horizontal,
@@ -91,6 +97,14 @@
 
 .participation-tile__title {
     margin-top: 0;
+
+    @media screen and (min-width: $breakpoint-md) {
+        font-size: $font-size-md;
+    }
+
+    @media screen and (min-width: $breakpoint-lg) {
+        font-size: $font-size-lg;
+    }
 }
 
 .participation-tile__btn {


apps/projects/templates/a4_candy_projects/includes/project_module_tile.html
```diff
@@ -5,7 +5,7 @@
     <a href="{{ module.get_absolute_url }}">
         <div class="participation-tile__type">
             <div class="participation-tile__content">
-                <h3 class="participation-tile__title">{{ module.name|truncatechars:100 }}</h3>
+                <h4 class="participation-tile__title">{{ module.name|truncatechars:100 }}</h4>
```
fuzzylogic2000, line 8: We cannot use the headlines for styling or font sizes as this will break a11y. Does this have the appropriate order of an h1, h2 and h3 before? Otherwise, you could add an utility class `u-h4-font-size` or something like that, that has the same sizing behaviour as an h4, but still use the h3 for the structure.

khamui: yes true. I restored it to h3 again and used the class, which is only used here to make the font-size change. That way I avoid to introduce a couple of utility classes (because of media queries).

Note: Anyway h2 is missing in the structure, but maybe because it is my dev-testing-content which does not contain proper descriptions etc.

fuzzylogic2000 (approved):
Nice! 
About the h3 that maybe shouldn't be there, we should probably check that. Could you open an issue and write down, what we have for headings on that (project detail) page and also which ones can get in there?

PR #1882, opened 2022-06-01
settings/base: add A4_USE_ORGANISATIONS_TERMS_OF_USE

Rineee:
Belongs to https://github.com/liqd/adhocracy4/pull/1107 and needs update of a4 hash after that is merged

### DIFF
--- a/adhocracy-plus/config/urls.py
+++ b/adhocracy-plus/config/urls.py
@@ -21,8 +21,6 @@ from adhocracy4.comments.api import CommentModerateSet
 from adhocracy4.comments_async.api import CommentViewSet
 from adhocracy4.follows.api import FollowViewSet
 from adhocracy4.polls.api import PollViewSet
-from adhocracy4.polls.api import VoteViewSet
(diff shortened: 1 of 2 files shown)


fuzzylogic2000 (commented):
I think this should go up a bit to the other a4 settings. Sorry, bit nitpicky today. :)

goapunk:
Looks good to me once moved up to the other A4_ settings. This means we need to remember to overwrite this in the pillar for single organization platforms before we release them (e.g. DIID), right?

adhocracy-plus/config/urls.py
```diff
@@ -67,9 +65,6 @@
 ct_router.register(r'comment-moderate', CommentModerateSet,
                    basename='comment-moderate')
 
-question_router = QuestionDefaultRouter()
-question_router.register(r'vote', VoteViewSet, basename='vote')
-
 urlpatterns = [
```
Rineee, line 68: We will also need to remove that in Opin and MeinBerlin, I will prepare pull requests, that we can then merge once we update a4 there..

PR #1894, opened 2022-06-07
assets/captcheck: add questions to tabbing order

phillimorland:


adhocracy-plus/templates/account/signup.html
```diff
@@ -52,7 +52,28 @@ <h1>{% trans "Register" %}</h1>
             <input type="hidden" name="{{ redirect_field_name }}" value="{{ redirect_field_value }}"/>
         {% endif %}
 
-        {% include 'a4_candy_contrib/includes/form_field.html' with field=form.captcha %}
+        {% with field=form.captcha %}
+        <div class="form-group">
+            <label for="{{ field.id_for_label }}">
+                {{ field.label }}{% if field.field.required %}<span role="presentation" title="{% trans 'This field is required' %}">*</span>{% endif %}
```
fuzzylogic2000, line 58: @phillimorland we did copy the wrong formfield template. But with this one (https://github.com/liqd/adhocracy-plus/blob/main/apps/contrib/templates/a4_candy_contrib/includes/form_field.html), the label is also there. :)

Is there another place where we also need to add the tabindex to the label? And if yes, could you point me to it?

phillimorland: also will fix and add to other form

apps/cms/contacts/templates/a4_candy_cms_contacts/form_field.html
```diff
@@ -24,7 +24,7 @@
   </div>
   {% else %}
   <div class="form-group">
-    <label class="mb-1">{{ field.help_text }} {% if field.field.required %} *{% endif %}</label>
+    <label class="mb-1" for="id_{{ field.name }}">{{ field.help_text }} {% if field.field.required %} *{% endif %}</label>
```
phillimorland, line 27: make contact form accessible

adhocracy-plus/templates/account/signup.html
```diff
@@ -52,7 +52,28 @@ <h1>{% trans "Register" %}</h1>
             <input type="hidden" name="{{ redirect_field_name }}" value="{{ redirect_field_value }}"/>
         {% endif %}
 
-        {% include 'a4_candy_contrib/includes/form_field.html' with field=form.captcha %}
```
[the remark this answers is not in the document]
phillimorland: @goapunk adding a tab index is a kinda hacky way to do this, not actually bad practice but we shouldn't 'need' to so adding it to all form fields wouldn't be ideal

apps/contrib/templates/a4_candy_contrib/includes/form_field.html
```diff
@@ -1,7 +1,7 @@
 {% load i18n widget_tweaks %}
 
 <div class="form-group">
-    <label for="{{ field.id_for_label }}">
+    <label for="{{ field.id_for_label }}" {% if tabindex is not None %}tabindex="{{ tabindex }}"{% endif %}>
```
phillimorland, line 4: @fuzzylogic2000 is this ok to do? it works but not sure if this is the place for logic, sorry will try get better with the django stuff!

fuzzylogic2000: Yes, that is a very good idea! Should have thought of that. :)
We could call the variable legal_tabindex to make more clear where it goes, but we can also change that when we need to add a tabindex elsewhere.
Should I do the commit squashing?

phillimorland: @fuzzylogic2000 yes please! thank you!

apps/captcha/assets/captcheck.js
```diff
@@ -122,7 +122,7 @@ window.onload = function () {
           const questionDiv = document.createElement('div')
           questionDiv.setAttribute('class', 'captcheck_label_message')
           questionDiv.setAttribute('id', 'captcheck_' + idp + '_label_message')
-          questionDiv.innerHTML = "<span class='captcheck_question_image' id='captcheck_" + idp + "_question_image'>" + data.question_i + "</span><span class='captcheck_question_access' id='captcheck_" + idp + "_question_access'>" + data.question_a + "</span><a href='' class='captcheck_alt_question_button' data-prefix='" + idp + "' id='captcheck_" + idp + "_alt_question_button' aria-role='button' aria-label='" + textLabel + "' tabindex='0'>" + textTextMode + '</a>'
+          questionDiv.innerHTML = "<span class='captcheck_question_image' tabindex='0' id='captcheck_" + idp + "_question_image'>" + data.question_i + "</span><span class='captcheck_question_access' tabindex='0' id='captcheck_" + idp + "_question_access'>" + data.question_a + "</span><a href='' class='captcheck_alt_question_button' data-prefix='" + idp + "' id='captcheck_" + idp + "_alt_question_button' aria-role='button' aria-label='" + textLabel + "' tabindex='0'>" + textTextMode + '</a>'
```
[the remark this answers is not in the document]
phillimorland: not really :(

PR #1897, opened 2022-06-09
apps/account: add users agreement on organisation terms of use to acc…

fuzzylogic2000:
…ount settings

Functionality is ok, 
still missing:

- [x] styling
- [x] org image
- [x] a11y
- [x] font awesome icon


phillimorland:
sure! lets check after standup

apps/account/templates/a4_candy_account/account_dashboard.html
```diff
@@ -43,6 +43,13 @@
                             {% trans 'Account Connections' %}
                         </a>
                     </li>
+                    <li class="dashboard-nav__page">
+                        <a href="{% url 'user_agreements' %}"
+                           class="dashboard-nav__item dashboard-nav__item--interactive {% if request.resolver_match.url_name == "user_agreements" %}is-active{% endif %}">
+                            <i class="far fa-square-check" aria-hidden="true"></i>
```
fuzzylogic2000, line 49: @khamui And a question for you: what is the deal with the a+ icons? We don't use font awesome? (I saw an issue about cleaning them up...) How do we get them?

phillimorland: was that the issue to clean up our custom icons? if so it was to create a clean set of custom icons with only the ones we needed but we pretty much only use them in interactive event

fuzzylogic2000: Yes, that issue is what made me think that we don't use font awesome. But then I only added it wrong and we do use font awesome?

fuzzylogic2000: So, we do use font awesome! But somehow, I could not find the right name. :)

apps/account/templates/a4_candy_account/user_agreements.html
```diff
@@ -0,0 +1,44 @@
...
+
+    <form method="post">
+        {% csrf_token %}
+
+        {{ formset.management_form }}
+        {% for agreement_form in formset.forms %}
+        <div class="d-flex bg--bottom-border">
+            <div class="pe-4">
+                {% if agreement_form.instance.organisation.logo %}
+                <img src="{% thumbnail agreement_form.instance.organisation.logo '0x160' %}" alt="" class="account__org-image" />
```
fuzzylogic2000, line 18: @phillimorland the image still shrinks in width, even after I added the width. I only managed to fix that if I wrap the image in a div. Is that ok to do? Or is there a class that makes an inline-block behave like a div? I tried `d-block`, but that did not work.

phillimorland: @fuzzylogic2000  in this case actually I think it's about flex item inheritance, the direct children of a flex parent can be manipulated but once there is a structural layer between flex has no effect so I think it's fine. (I hope that's what's happening :p)

fuzzylogic2000: @phillimorland Ah, so I have to add the div to make the image not shrink? But why does the div without the image, but only blue bg not shrink?

apps/account/templates/a4_candy_account/account_dashboard.html
phillimorland: yes we still use it, closest i could find is <i class="fas fa-check-square"></i> we haven't updated to 6 yet I don't think

apps/account/templates/a4_candy_account/user_agreements.html
phillimorland: oh that I don't know, does it still do it when we use a normal image with display: block and height and width, not a thumbnail tag? asking as the inline styles will always override the loaded ones so maybe that's why?

phillimorland: also you can try flex-shrink: 0 on the image

fuzzylogic2000: adding the div was the only way, I tried everything else. So, here it is. Another div around the image, which was in an extra div anyway. :woman_shrugging:

phillimorland (commented):
Ok I had a little play due to the background img and came up with the below possibility to ensure background colour is shown when no logo, also this way we remove the extra div class name `pe-4 flex-shrink-0`:
`.account__org-image {
    height: 2 * $spacer;
    max-width: 2 * $spacer;
    margin-right: 0.5 * $spacer;
    flex: 0 0 (2 * $spacer);
}`
it is setting grow and shrink to 0 however flex items are kinda comparative so we also set a max width and initial width to ensure it doesn't change.
Sorry I didn't think of this yesterday!

Rineee (approved):
Awesome! Works like a charm!

adhocracy-plus/assets/scss/components/_background.scss
```diff
@@ -12,3 +12,9 @@
     border-top: 2px solid $gray-lightest;
     margin-top: $spacer;
 }
+
+.bg--bottom-border {
+    padding-top: 2 * $spacer;
+    padding-bottom: 2 * $spacer;
+    border-bottom: 2px solid $gray-lightest;
+}
```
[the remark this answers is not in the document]
fuzzylogic2000: Yes. Why is it not in the same place as the other divider? And do we need the other one, which is also only used somewhere in the account settings, then? Will try that!

adhocracy-plus/assets/scss/components/_account.scss
```diff
@@ -0,0 +1,8 @@
+.account__org-image {
+    height: 2 * $spacer;
+    width: 2 * $spacer;
+}
+
+.account__org-image--bg {
+    background-color: $brand-primary;
+}
```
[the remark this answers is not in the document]
fuzzylogic2000: It should be, it is on the div in the else. 
But I will try with the other stuff you suggested and put the if directly into the image tag classes.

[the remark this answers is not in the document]
phillimorland: let me know if I can help

phillimorland (approved):
fab! :)

PR #1899, opened 2022-06-09
Kl 2022 06 terms of use issues

fuzzylogic2000:


fuzzylogic2000:
I just thought, that we can probably make the user and organisation not editable directly in the model... So, have to try before we merge second commit like that.

fuzzylogic2000:
> I just thought, that we can probably make the user and organisation not editable directly in the model... So, have to try before we merge second commit like that.

Alright, I removed that from the admin again. In the other PR #1897 , I make the fields not editable in the  model. But as the add-button in the django-admin is then still broken, I kept that the add-button is removed.

Rineee (approved):
Nice!

PR #1901, opened 2022-06-09
js/idea_remarks.js: not working jquery selector replaced and simplify…

### LINKED ISSUE #1843: Idea remark broken
https://aplus-dev.liqd.net/phillitopictest-org/ideas/2022-00604/
~~idea remark will not save~~
**behaviour:** moderator remark appears not to save when clicking save, box does not close, green tick does not appear, user must refresh the page in order to check if it is saved and see the green tick
**expected behaviour:** on clicking save the moderator remark box closes and a green tick appears on the button, no page refresh needed 

Screenshot?

khamui:
… to match expected behavior

fixes #1843

### DIFF
--- a/apps/moderatorremark/assets/idea_remarks.js
+++ b/apps/moderatorremark/assets/idea_remarks.js
@@ -1,3 +1,4 @@
+import bootstrap from 'bootstrap'
 const $ = require('jquery')
 const a4api = require('adhocracy4').api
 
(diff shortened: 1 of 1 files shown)


phillimorland (changes requested):
cool, it now closes on save however the green tick does not appear. Element was originally written for mB and there the green tick behaviour still works, just not the closing on save, maybe see how tick behaviour is achieved there?
https://meinberlin-dev.liqd.net/mapideas/2022-00695/

khamui:
oops...! should be fixed now @phillimorland

phillimorland (approved):
Perfect! :) Quick question, if we refactor lines 21 to 24 do we still need jquery?

PR #1902, opened 2022-06-09
apps/dashboard_project/richtext collapsible: fixes #1811

### LINKED ISSUE #1811: a+: list in collapsible element not shown
**URL:** https://aplus-dev.liqd.net/frede-test/dashboard/projects/titel-ihres-projekts-2/information/
**user:** all
**expected behaviour:** when i add a list in a collapsible element the list is shown correctly
**behaviour:** The list is not shown in a collapsible element
**important screensize:**
**device & browser:** Tested on macOS 12.3 with latest Safari and Chrome
**Comment/Question:** 

Screenshot?
Bildschirmfoto 2022-04-26 um 14 31 45
Bildschirmfoto 2022-04-26 um 14 32 10

khamui:
fixes #1811

### DIFF
--- a/adhocracy-plus/assets/scss/components/_accordion.scss
+++ b/adhocracy-plus/assets/scss/components/_accordion.scss
@@ -40,12 +40,6 @@
 .accordion__body {
     padding: $padding;
 
-    ul {
-        list-style: none;
-        margin-left: 0;
-        padding: 0;
-    }
-
     li {
         margin-bottom: 0.2em;
     }


adhocracy-plus/assets/scss/components/_blocks.scss
```diff
@@ -286,3 +286,8 @@ $block-padding-md: 7.5 * $padding;
 .block-quote__author {
     font-family: $font-family-serif;
 }
+
+.accordion__body ul {
+    list-style: initial !important;
+    padding: 0 1rem !important;
+}
```
phillimorland, line 293: This should be in the accordion component file. We actually already set the styling in there to none, did you check both richtext accordions? (wagtail and ckeditor) I assume we did that for a reason so would test to check if reason is due to design breakages.
code line:
https://github.com/liqd/adhocracy-plus/blob/ddd1853a412abb91748046db4837b3ffb9eb6661/adhocracy-plus/assets/scss/components/_accordion.scss#L43

phillimorland (changes requested):
see comment

adhocracy-plus/assets/scss/components/_blocks.scss
khamui: @phillimorland thanks for the hint! I removed this reset, because i cannot see why it is there. This way it is working fine for wagtail (faq blocks) and ckeditor. Can't see in code, why it was reset though.

phillimorland (approved):
cool!

PR #1904, opened 2022-06-13
apps/signup/social: fix radio button width scale

khamui:


phillimorland (approved):
sweet and great keeping complib update! :)

PR #1905, opened 2022-06-13
config/settings/organisation logo: max_resolution

khamui:
a4 already updated in this branch, can be merged like that


Rineee:
Yes, it works! But could you also add the max 800x800 to the helptext? You can find it here: https://github.com/liqd/adhocracy-plus/blob/6228d6679d0d494d2c847dbd300af985910a4130/apps/organisations/models.py#L67. Will require a migration though.

goapunk (approved):
cool!

PR #1915, opened 2022-06-14
ci: use node 18

goapunk:


phillimorland:
@goapunk should we review this or just test it? not sure the purpose of assignment is. It works for me fyi.

goapunk:
review and merge it I'd say

PR #1920, opened 2022-06-14
apps/account: test account user agreements view

fuzzylogic2000:


fuzzylogic2000:
actually I will add some more tests!

fuzzylogic2000:
@goapunk Let's merge, then?! I think without @Rineee noone else will have a look at tests. Or do you want to have a look @phillimorland ? There is also the error message change for the emails in here.

Issue #1924, opened 2022-06-15
a+ | Captcha a11y user story #6038


fredewe:
**URL:** https://aplus-dev.liqd.net/accounts/signup/?next=/
**user:** unregistered user with activated screen reader
**expected behaviour:** The screen reader reads all relevant information i need to register on adhocracy+
**behaviour:** Not all relevant things are read by the screen reader, there is information missing or the SR does not read it like we want it 
**important screensize:**-
**device & browser:** MacOS and Safari using MacOS VoiceOver and Chrome using Screen Reader extension 
**Comment/Question:** 
I don't know how to deal with this because i don't know if i can test it the way a person using SR would like it to be.
There is a screenshot where i included the thinks i have noticed.
Screenshot?
![Bildschirmfoto 2022-06-15 um 16 11 38](https://user-images.githubusercontent.com/5470479/173851605-10f3c4b3-5294-4c25-ba2d-96dfc7721316.png)

phillimorland:
@fredewe The image version is not accessible for SR only text version, adding images names in the code means that bot just has to word match which is very simple also would mean we would have to use a different captcha which was not within the points and as far as opensource captcha's go, this is already one of the better ones.

The helptext is not tabbable as it is not the form label, it is only readable, this is a different function on SR's, normally `alt shift arrow`. It should however be focusable when there is a form error which is not a functionality we have anywhere and appears to be very difficult in django, there is a solution here which maybe @fuzzylogic2000 or @Rineee can check? my very quick attempt at implementation failed.

The not reading it - as math is definitely a problem and i'll check it,

Let me know if anything wasn't clear or if you have any questions.

fredewe:
@phillimorland thank you very much for clarification, that helped me a lot understanding the possibilities we have here! If the reading as math is somehow possible it would be great.

phillimorland:
further conversation about django forms and a11y here https://groups.google.com/g/django-updates/c/eRJbEvjka6U

phillimorland:
@fredewe just noticed the other issue about the links not being read, that's not an issue I've come across with any of my screen readers, can you check are they also not tabbed to?

phillimorland:
NOTE: link reading is only an issue in safari with voiceover, works in ff and chrome

fredewe:
@phillimorland Thanks for investigating! That is fine for me!

PR #1925, opened 2022-06-16
templates/* add role alert to our error messages and aria-invalid acc…

phillimorland:
…ording to WCAG guidlines

kind of related to #1924 but not a fix as such


apps/contrib/templates/a4_candy_contrib/includes/form_field.html
```diff
@@ -8,15 +8,21 @@
...
+    <p class="form-hint" id="{{ field.id_for_label }}_hint">
         {{ field.help_text }}
-    </div>
+    </p>
     {% endif %}
     {% endblock %}
     {% block field %}
         <div class="widget widget--{{ field|widget_type }}">
+          {% if field.errors %}
+            {{ field|attr:"aria-invalid=true" }}
```
phillimorland, line 19: it would be good to also have `aria-describedby="{{field.id_for_labe }}_hint` but wasn't sure how

apps/contrib/templates/a4_candy_contrib/includes/form_field.html
```diff
@@ -8,15 +8,21 @@
...
         {{ field.help_text }}
-    </div>
+    </p>
     {% endif %}
     {% endblock %}
     {% block field %}
         <div class="widget widget--{{ field|widget_type }}">
+          {% if field.errors %}
+            {% render_field field|attr:"aria-invalid=true"|attr:"aria-describedby=hint_{{field.id_for_label}}" %}
+          {% else %}
```
phillimorland, line 20: I would sooooooo love this to work (`aria-describedby`) but sadly it doesn't, I have been faffing for too long though so will see if you guys have any suggestions and if not will remove

goapunk: @phillimorland I think this works, maybe you can double check?

goapunk: alternatively we can add a templatetag I think

phillimorland: oooo nice, I tried with `with``but didn't use/find out about `|add`, will check it :)

apps/users/templatetags/userindicator.py
```diff
@@ -29,3 +29,8 @@ def get_next_url(request):
 def userindicator(context):
     context['redirect_field_value'] = get_next_url(context['request'])
     return context
+
+
+@register.filter(name="get_key_value")
+def get_key_value(some_dict, key):
+    return some_dict.get(key, '')
```
fuzzylogic2000, line 36: So, we can remove this templatetag again, right?

fuzzylogic2000 (commented):
I could not find a formfield with a helptext where I could produce an error apart from the images and I could not find the describedby in there. Could you give me a hint where to look?
Everything I could test is great! 
But let's squash the commits and remove the templatetag before merge.

apps/users/templatetags/userindicator.py
phillimorland: and yes that was my other attempt at adding it, forgot to remove, I can squash commit if you think it's fine to do it like this

fuzzylogic2000 (approved):
Supernice!

PR #1928, opened 2022-06-17
Kl 2022 06 agreement inconsistencies

fuzzylogic2000:


apps/organisations/mixins.py
```diff
@@ -64,3 +72,12 @@ def _get_terms_of_use_label(self):
...
         return mark_safe(label)
+
+
+class UserFormViewMixin(FormView):
+    """Adds the user from the request to the form_kwargs."""
+
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        kwargs['user'] = self.request.user
+        return kwargs
```
fuzzylogic2000, line 83: I went a bit back and forth about where to put this. Instead of adding this one mixin, we could also add it to every of the 6 views separately or add two extra views inheriting from AbstractIdeaCreateView and AbstractIdeaUpdateView. In the end I went for this approach because it is the shortest one and put it here with the other mixin. But I am also happy to put it elsewhere when that makes more sense.

phillimorland: I think keeping all the mixins together is good, unless i'm missing something djangoy

phillimorland (approved):
nice, works and think mixin solution is elegant

PR #1929, opened 2022-06-20
assets/captcheck: update to logical tabbing order

phillimorland:


goapunk (approved):
works fine, looks a bit uglier though :)

PR #1932, opened 2022-06-20
Jd 2022 06 remove webpack library export

goapunk:


adhocracy-plus/assets/js/unload_warning.js
```diff
@@ -26,11 +28,11 @@ $(function () {
   //   })
   // }
```
phillimorland, line 29: Now the embed is gone can we add this function back in?

phillimorland (commented):
looks good and seems to work with uploads and category adding so should be all good, just couple comments

adhocracy-plus/assets/js/unload_warning.js
```diff
@@ -3,16 +3,18 @@
...
     if (target.includes('dashboardToggle')) {
       submitted = true
     }
-    $(window).on('beforeunload', function (e) {
+    window.addEventListener('beforeunload', (e) => {
       if (!submitted) {
-        const string = django.gettext('If you leave this page changes you made will not be saved.')
+        // string is ignored on most modern browsers
+        const string = django.gettext('If you leave this page changes you made will not be saved. WTF')
+        e.preventDefault()
```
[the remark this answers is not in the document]
goapunk: ups :D

adhocracy-plus/assets/js/unload_warning.js
goapunk: technically the embed isn't gone yet, is it?

goapunk: uncommented it, seems to work

phillimorland (approved):
very nice!

PR #1933, opened 2022-06-20
eslint cleanup

khamui:
…one react linting error + disable proptypes rule


package.json
```diff
@@ -62,14 +62,9 @@
     "eslint": "8.17.0",
     "eslint-config-standard": "17.0.0",
     "eslint-config-standard-jsx": "11.0.0",
-    "eslint-plugin-import": "2.26.0",
     "eslint-plugin-jest": "26.5.3",
     "eslint-plugin-jsx-a11y": "6.5.1",
-    "eslint-plugin-n": "15.2.2",
-    "eslint-plugin-node": "11.1.0",
-    "eslint-plugin-promise": "6.0.0",
```
goapunk, line 70: should we remove this one ? or actually activate it? for the latter we'll get some errors we have to fix

package.json
```diff
@@ -62,14 +62,9 @@
     "eslint": "8.17.0",
     "eslint-config-standard": "17.0.0",
     "eslint-config-standard-jsx": "11.0.0",
-    "eslint-plugin-import": "2.26.0",
```
goapunk, line 65: same here: activate or remove? or is it unnecessary?

package.json
```diff
@@ -62,14 +62,9 @@
...
     "eslint-plugin-jest": "26.5.3",
     "eslint-plugin-jsx-a11y": "6.5.1",
-    "eslint-plugin-n": "15.2.2",
```
goapunk, line 68: same here: do we want to have this? if yes we should activate it instead of removing

package.json
khamui: since we didn't use it (at least recently) i would remove them to reduce the deps footprint. some of them even have the same rules. What do you think @phillimorland ?

package.json
khamui: this might be helpful (removed because we didnt activate it, but will do).

.eslintrc
```diff
@@ -18,12 +19,19 @@
...
+    "plugin:n/recommended"
   ],
   "rules": {
     "jsx-quotes": [2, "prefer-double"],
     "jsx-a11y/no-onchange": "off",
-    "react/prop-types": "off"
+    "react/prop-types": "off",
+    "n/no-missing-require": "off",
+    "n/no-unsupported-features/es-syntax": "off",
+    "n/no-missing-import": "off",
```
khamui, line 32: linting regarding imports is done better by the other package we use (e.g. `eslint-plugin-import`). So disabling these rules.

adhocracy-plus/assets/js/app.js
```diff
@@ -20,7 +20,7 @@ import {
 } from 'adhocracy4'
 
 import * as ReactDocuments from '../../../apps/documents/assets/react_documents.jsx'
-import * as ReactInteractiveEvents from 'a4_candy_interactive_events'
+import * as ReactInteractiveEvents from 'a4_candy_interactive_events' // eslint-disable-line
```
phillimorland, line 23: why disabled here?

package.json
phillimorland: as we have n I think we should remove this one

phillimorland (commented):
cool, less fixing then I thought which is nice, I wonder if we should look at our app js just because of the disabled line, maybe the apps should all have separate entry points instead of being initialised there? what do you guys think @khamui @goapunk ?

khamui:
> 

sounds good with the entry points thing? i will have a look if i get on top of it? Idea is to have it set in webpack instead of requiring them in app.js right?

.eslintrc
```diff
@@ -44,11 +44,15 @@
     "static/"
   ],
   "settings": {
+    "import/core-modules": [ django ],
     "import/resolver": {
       "node": {
         "extensions": [".js", ".jsx"]
+      },
+      "webpack": {
```
khamui, line 52: from this line on, this differs from .eslintrc in a4. because there we do not have webpack and obviously cannot ignore adhocracy4.

adhocracy-plus/assets/js/app.js
```diff
@@ -19,9 +19,9 @@ import {
   widget as ReactWidget
 } from 'adhocracy4'
 
-import * as ReactDocuments from '../../../apps/documents/assets/react_documents.jsx'
-import * as ReactInteractiveEvents from 'a4_candy_interactive_events'
-import * as ReactLanguageChoice from '../../../apps/organisations/assets/react_language_choice.jsx'
+import { renderDocumentManagement } from 'react_documents'
```
goapunk, line 22: is there any advantage using the webpack resolve things over just directly importing it? Needing the webpack resolve stuff feels a bit heavy / more to maintain than just having the `../../../apps/` import

goapunk (approved):
cool, looks like a nice change! only added one comment regarding the webpack resolve

adhocracy-plus/assets/js/app.js
khamui: not really thought about performance.
I thought it is less to maintain if we would for instance use those apps in different projects, but having them each in different folder structures. And we only set them in one spot (webpack config), rather than having different app.js files. But i might idealize it here...

personally i do not really like `../../` maybe i can try resolve the path into `apps/` and then we have something like:
`apps/organisations/assets/react_language_choice.jsx` for instance?

goapunk: it's not about performance, I just wasn't sure if having code in different files (e.g. code for imports in webpack config) makes it easier or harder to maintain. I also don't know it, so if you think we should stick to your solution I'm absolutely fine with that :) 
 > I thought it is less to maintain if we would for instance use those apps in different projects, but having them each in different folder structures. And we only set them in one spot (webpack config), rather than having different app.js files. But i might idealize it here...

I didn't really take that into account so maybe you are right

khamui: @goapunk i decided to go without webpack resolving, because apart from maintainance that might be more complicated indeed, this way i can keep `.eslint` identical in a4 and aplus. So latest commit has it already.

package.json
```diff
@@ -62,14 +62,13 @@
     "eslint": "8.18.0",
     "eslint-config-standard": "17.0.0",
     "eslint-config-standard-jsx": "11.0.0",
+    "eslint-import-resolver-webpack": "0.13.2",
```
goapunk, line 65: do we still need that without the webpack ? :)

khamui: wahh

good catch thanks, removed

PR #1941, opened 2022-06-27
Revert "webpack: don't export entry points as library"

goapunk:
This reverts commit 5745d5c59204d19dbbeb090c71fde5d50a5f8651.


Rineee (approved):
Yay, polls are back!

PR #1949, opened 2022-06-28
accordion: add samesite attribute

goapunk:


khamui:
while checking this, seems the accordion behaviour is broken. Probably not related to this commit, though.
shall i create an issue? Or do you want to have a look into that?

PR #1953, opened 2022-06-28
projects/serializers: sending project.information html formatted

khamui:


goapunk (changes requested):
the serializer tests now fail and need fixing.

tests/projects/test_app_project_api.py
```diff
@@ -89,10 +89,11 @@ def test_app_project_api_single_poll_module(
 @pytest.mark.django_db
 def test_app_project_serializer(project_factory, module_factory, phase_factory,
                                 apiclient):
+    html_string = '<p>text with a <strong>bold</strong> bit</p>'
```
fuzzylogic2000, line 92: Even though, we do not remove the tags anymore, you could test that the strip works and removes whitespace.

fuzzylogic2000 (commented):
Nice! Could you now also squash the commits? (When you just do a fixup with the second one, you will not have the weird commit message with "squash" in it.)

khamui:
@fuzzylogic2000 oh makes sense. Now it tests the strip of whitespaces

PR #1954, opened 2022-06-28
Ks 2022 06 use blueprint types

Rineee:


apps/dashboard/blueprints.py
```diff
@@ -10,6 +11,14 @@
...
 from apps.topicprio import phases as topicprio_phases
 
+# this overwrites adhocracy4.dashboard.blueprints.ProjectBlueprint in order to
+# add the blueprint type
+ProjectBlueprint = namedtuple(
+    'ProjectBlueprint', [
+        'title', 'description', 'content', 'image', 'settings_model', 'type'
+    ]
+)
+
```
Rineee, line 21: Not sure about adding it here, as we are eventually also going to add it to other projects? We could also add another ProjectBlueprintWithType or sth to A4?Or ModuleBlueprint :stuck_out_tongue_winking_eye:  @fuzzylogic2000

fuzzylogic2000: Can we also add it and just not use (leave out) in other projects? Or does it have to have a type as soon as we add it? Or only add the type when the settings are there?
Otherwise, I am also fine with having a ProjectBlueprintWithType in a4. :)

fuzzylogic2000 (approved):
Nice!

apps/dashboard/blueprints.py
fuzzylogic2000: :( Well, then I don't really have an opinion. Whatever you think is better is fine with me!

Rineee: Same here :) But then I will leave it here and we can move it once we use it somewhere else..

fuzzylogic2000 (commented):
Perfect! Can you squash the first three commits when you update the a4 version, though?

fuzzylogic2000 (approved):
Cool!

PR #1961, opened 2022-07-06
replace deprecated BASE_URL with WAGTAILADMIN_BASE_URL

goapunk:


fuzzylogic2000 (commented):
I don't get if that setting should be the base urls or the base url of the wagtail admin? The docs sound like we could just replace it, but still?

fuzzylogic2000 (approved):
Well in the docs it days that it should be the base, which is all fine, then.
Thank you for fixing!

PR #1962, opened 2022-07-06
sending organisation_logo to frontend

khamui:


Rineee (changes requested):
Cool! But I get an error if no logo is there, so you probably need to check if organistaion.log: return url, else return None or sth like that

khamui:
oh yes. Thanks! just added what you requested

Rineee (approved):
Yeah!

PR #1963, opened 2022-07-06
unload_warning: fix ckeditor pop-up error

goapunk:
the unload warning currently fails when you try to enter a url for the media embed. We don't need the warning for the pop-up anyway, so just skip it


khamui:
cool! yes no need for the warning in that case

PR #1965, opened 2022-07-07
deps: update and tag a4

fuzzylogic2000:


Rineee (approved):
Cool!

Rineee:
Should I merge this now or do we wait for possible A4 comment changes and then merge with updated tag @fuzzylogic2000?

fuzzylogic2000:
@Rineee I don't know. Uh, or maybe, I remove the tag again and replace it by the hash, so that we can then use the tag when we actually finished a4? Yes, will do that now! (Or soon :p)

PR #1966, opened 2022-07-11
package.json: add required packages back, ChapterNavItem: update impo…

phillimorland:
…rt to work


fuzzylogic2000 (approved):
Nice!

goapunk (approved):
cool! I wonder where this is coming from though. I checked back till January and the package isn't there.

PR #1989, opened 2022-07-19
docs/api: add some curl commands

fuzzylogic2000:
As I always forget how to do this, I added some docs, we can add to over time.


Rineee (approved):
Really good idea!! I actually have a file with more commands, will add it to the docs then!

PR #1992, opened 2022-07-21
cms: form_data already is json in wagtail 3.x

goapunk:


phillimorland (approved):
LGTM will test on dev

PR #1993, opened 2022-07-22
projects/api: only allow authenticated users

Rineee:


fuzzylogic2000 (approved):
Nice! But merging needs to wait for the other PR, right?

goapunk (approved):
cool!

Rineee:
> Nice! But merging needs to wait for the other PR, right?

Yes, they should be merged together!

PR #1994, opened 2022-07-25
account//config/urls: add account api and viewset

Rineee:
This allows PUT or GET on api/account/ and operates on the logged in user, if no authenticated user, returns 401.


fuzzylogic2000 (approved):
Supercool! Works in the browser and with curl.

PR #1995, opened 2022-07-25
[6233] apps/users: add user API to get user info in app

fuzzylogic2000:


apps/users/serializers.py
```diff
@@ -0,0 +1,18 @@
...
+
+from .models import User
+
+
+class UserSerializer(serializers.ModelSerializer):
+    is_self = serializers.SerializerMethodField()
+
+    class Meta:
+        model = User
+        fields = ('pk', 'username', '_avatar', 'is_self')
```
Rineee, line 11: Do we need the pk?

Rineee (approved):
Very nice! One little question, but can be merged!

apps/users/serializers.py
fuzzylogic2000: No, not anymore! Will remove it!

PR #2001, opened 2022-07-26
config/urls: remove name of api/account

Rineee:
This was interfering with https://github.com/liqd/adhocracy-plus/blob/1b9c351143aeb87f97f5b805ba2e3d21e8c8ab80/apps/account/urls.py#L8 and now the link to account settings lead to api :see_no_evil:

### DIFF
--- a/adhocracy-plus/config/urls.py
+++ b/adhocracy-plus/config/urls.py
@@ -89,7 +89,7 @@ urlpatterns = [
     re_path(r'^api/', include(likes_router.urls)),
     re_path(r'^api/', include(router.urls)),
     re_path(r'^api/login', obtain_auth_token, name='api-login'),
-    re_path(r'^api/account/', AccountViewSet.as_view(), name='account'),
+    re_path(r'^api/account/', AccountViewSet.as_view()),
     re_path(r'^upload/', user_is_project_admin(ck_views.upload),
             name='ckeditor_upload'),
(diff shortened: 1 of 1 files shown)


adhocracy-plus/config/urls.py
```diff
@@ -89,7 +89,7 @@
     re_path(r'^api/', include(likes_router.urls)),
     re_path(r'^api/', include(router.urls)),
     re_path(r'^api/login', obtain_auth_token, name='api-login'),
-    re_path(r'^api/account/', AccountViewSet.as_view(), name='account'),
+    re_path(r'^api/account/', AccountViewSet.as_view()),
```
fuzzylogic2000, line 92: Upsi, right! You could name it api-account, though?! If we want a name.

fuzzylogic2000 (approved):

adhocracy-plus/config/urls.py
Rineee: Did that first, but then thought we dont use the name anyway, so removed it?

Rineee: Do you want it back? :)

fuzzylogic2000: No, I don't care. Where do we normally use it, though? In the tests? But anyway, can add it when we need it!

Rineee: Ah yes, we might need it for the tests, right! Well, too late :stuck_out_tongue_closed_eyes:

PR #2003, opened 2022-07-29
ideas/api: do not copy request.data, but send image deletion info

Rineee:


fuzzylogic2000 (commented):
Looks very good! I did not try it out, though. :)

phillimorland (approved):
looking good, merging to test

PR #2015, opened 2022-08-02
assets/images: add png fallback images

phillimorland:


phillimorland:
@Rineee not sure how far you got here but here are the images

PR #2017, opened 2022-08-02
Jd 2022 08 local a4

goapunk:
add makefile option to easily set a4 to local in package.json and requirements/base. Needs the a4 repo to be in the parent folder / `../adhocracy4`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,7 @@ help:
 	@echo "  make start-postgres  				-- start the local postgres cluster"
 	@echo "  make stop-postgres   				-- stops the local postgres cluster"
 	@echo "  make create-postgres   			-- create the local postgres cluster (only works on ubuntu 20.04)"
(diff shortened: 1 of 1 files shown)


Rineee:
Not sure I get it.. so I do make local-a4, but then I need to make install after, right? But then, when I add changes in a4, I need to do that again to show them, right? Is there also a way to have a command for the npm link thing, so changes in a4 are updated?

goapunk:
> Not sure I get it.. so I do make local-a4, but then I need to make install after, right? But then, when I add changes in a4, I need to do that again to show them? Is there also a way to have a command for the npm link thing, so changes in a4 are updated?

arg, right, I forgot about the `npm link`. I kinda just did it this way because you need to do it for the python part anyway. I'll add `make install` and `npm link`

fuzzylogic2000 (approved):
Nice!

PR #2018, opened 2022-08-04
Ks 2022 08 add images app user serializers

Rineee:
The update a4 commit is to be able to show images and fallback images in comments in the app.

So depends on https://github.com/liqd/adhocracy4/pull/1176 and **hash needs update after that is merged**.


fuzzylogic2000 (approved):
Works great in a+!
So, if it is also good in the app, please merge @khamui !

goapunk (approved):
looks good to me

PR #2081, opened 2022-11-24
[#6750] a4dashboard: add communication form

phillimorland:
The last commit depends on https://github.com/liqd/adhocracy4/pull/1289, so that need to be merged first and the hash added.

There are still some things that need to be done:
- checkboxes for logos
- image creation with input data and for different templates
- preview/download of generated image
- tests

But maybe this is a state that could be merged and we can on from there?

### DIFF
--- a/adhocracy-plus/templates/a4images/image_upload_widget_simple.html
+++ b/adhocracy-plus/templates/a4images/image_upload_widget_simple.html
@@ -1,0 +1,30 @@
+{% load i18n %}
+<div class="upload-wrapper form-control-upload">
+    <div class="upload-wrapper__fields">
+        {{ file_input }}
+        {{ text_input }}
+        <div class="upload-wrapper__action-parent">
+        {% if has_image_set and not is_required %}
+            <label for="{{ checkbox_id }}"
+                class="upload-wrapper__action btn btn--danger"
+                title="{% trans 'Remove the picture' %}">
+                <i class="far fa-trash-alt" aria-label="{% trans 'Remove the picture' %}"></i>
+            </label>
+        {% else %}
+            <label for="{{ id }}"
+                class="upload-wrapper__action btn btn--primary-filled"
+                title="{% trans 'Upload a picture' %}">
+                <i class="fas fa-cloud-upload-alt" aria-label="{% trans 'Upload a picture' %}"></i>
+            </label>
+        {% endif %}
+        </div>
+    </div>
+    <div class="form-{{ id }} upload-wrapper__preview col-sm-3">
+        {{ checkbox_input }}
+        <img id="{{ file_id }}"
+             src="{% if has_image_set %}{{ file_url }}{% else %}data:image/gif;base64,R0lGODlhAQABAAD/ACwAAAAAAQABAAACADs={% endif %}"
+             data-upload-preview="{{ id }}"
+             alt=""
+        >
+    </div>
+</div>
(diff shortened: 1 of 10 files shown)


phillimorland:
@fuzzylogic2000 didn't get far, actually not sure if I put it in right spot but here you go, good luck! :)

Rineee:
@fuzzylogic2000 Also did not get super far, added everything in a separate commit, but can probably be squashed with the first. For now I just added a basic form for the project, so not with links, not sure this is the way to go.. feel free to continue/change, if you dont get there, I will continue on Tuesday!

apps/organisations/forms.py
```diff
@@ -161,3 +190,73 @@ def __init__(self, *args, **kwargs):
...
+
+class CommunicationContentCreationForm(forms.Form):
+
+    def __init__(self, project=None, format=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        sizes = SOCIAL_MEDIA_SIZES[format]
+
+        self.fields['title'] = forms.CharField(
+            max_length=sizes['title_max_length'],
```
fuzzylogic2000: That's a good idea, with the sizes thingy!

apps/organisations/forms.py
```diff
@@ -253,10 +257,42 @@ def __init__(self, project=None, format=None, *args, **kwargs):
...
-
+                        format(self.sizes['img_min_width'],
+                               self.sizes['img_min_height'])),
         )
 
         if project:
             self.fields['title'].initial = project.name
             self.fields['description'].initial = project.description
+
+    def clean_image(self):
```
fuzzylogic2000, line 268: This did not get in here from the a4 image stuff, right? I mean, it did not work with the sizes given here dynamically.

adhocracy-plus/templates/a4images/image_upload_widget_simple.html
```diff
@@ -0,0 +1,30 @@
...
+                title="{% trans 'Upload a picture' %}">
+                <i class="fas fa-cloud-upload-alt" aria-label="{% trans 'Upload a picture' %}"></i>
+            </label>
+        {% endif %}
+        </div>
+    </div>
+    <div class="form-{{ id }} upload-wrapper__preview col-sm-3">
+        {{ checkbox_input }}
+        <img id="{{ file_id }}"
+             src="{% if has_image_set %}{{ file_url }}{% else %}data:image/gif;base64,R0lGODlhAQABAAD/ACwAAAAAAQABAAACADs={% endif %}"
```
fuzzylogic2000, line 25: What does the else do?

fuzzylogic2000 (approved):
I do have questions, but it works great and we should try and then improve, so I will merge!

apps/organisations/forms.py
Rineee: This was the thing that I could not make work otherwise.. I tried to use the validate_image function from a4 images and just add that to the image fields validators, but that did not give back the size/height validation error, but 'image is corrupt', so I gave up on that and did it this way.

Rineee: Does that answer the question?

adhocracy-plus/templates/a4images/image_upload_widget_simple.html
Rineee: This is just a complete copy of the other image widget template, but without the info box saying 'your image will be uploaded once you clicked save'. So, I did not write this code, but I think the else is the case when the image has not been saved yet and just showing the preview of the image in memory.

PR #2104, opened 2022-11-28
[updates] deps react18

phillimorland:


fuzzylogic2000 (approved):
Nice! The React componnets, I looked at all worked!
We wouldn't have needed the affiliation fix here, I think they are required!? But it's also good to have in case we do make them unrequired again.

fuzzylogic2000:
This needs a rebase, but then it's good to be merged!

phillimorland:
@fuzzylogic2000 will wait for tests and merge then :)

Issue #2103, opened 2022-11-28
styling issues from updates


fuzzylogic2000:
~~- [ ] "Save"-button in dashboard is not wide enough~~ (same on stage)
~~- [ ] blue of dashboard publish progress line seems to be wrong~~ (same on stage)
- [x] donation banner on LIQD (wagtail)-pages does not have background colour and disappers from time to time (depending on background of pages)
- [x] jsx errorlist text black not white
- [x] text commenting (documents) chapter btns wrong
- [x] text commenting (documents) structure wrong (nested rows?)

_Originally posted by @fuzzylogic2000 in https://github.com/liqd/adhocracy-plus/pull/2099#pullrequestreview-1195683160_

lthueer:
Hi @phillimorland I had a quick look and noticed the following:
- wagtail buttons have only white fonts (before it was white + black depending on background)
- language dropdown grey not black
- create new project white background
- white font on buttons in dashboard
- selecting date, missing black font + time boxes look off
- add module has blue font (not black)
- image upload symbol background not there (or maybe that's new?)
- light blue in background colour seems to be different
- turquoise seems to be slightly different? (user profile dropdown)
- save button blue outline missing + when selected turn dark blue
- donation banner has turquoise background
- finished project in dashboard project overview has wrong turquoise & private/semi-public has black colour on black background
- background grey copyright for images darker than on prod
- in general there seem to be a problem with black and white font (e.g. project header, category labels
- send comment (idea detail page) or idea button no background colour

jostertag:
- [ ] calender symbol/button in phases (dashboard) with white font instead of black
- [ ] add category, add label and export excel buttons in dashboard all have blue font (not black)
- [ ] You cannot see the card anymore (its just grey) --> I don't know if its related to styling issues
- [ ] labels (as merkmale, category, budget) in proposals/ideas with black font (should be white font)

lthueer:
@phillimorland I noticed the following two points while testing:
- [x] font colour a+ startpage, font colour doesnot adapt to background: https://aplus-dev.liqd.net/
- [ ] dashboard light blue looks different

phillimorland:
@jostertag what do you mean by card?

jostertag:
@phillimorland I meant map! When I tested I couldn't see it anymore, but this is no issue anymore, so I deleted it. Sorry for the confusion.

phillimorland:
these all seem to be done

PR #2107, opened 2022-11-29
[updates/issue] fixing styling issues from rm planifolia 

phillimorland:
partial fix for #2103

### DIFF
--- a/adhocracy-plus/assets/scss/components/_button.scss
+++ b/adhocracy-plus/assets/scss/components/_button.scss
@@ -4,27 +4,25 @@
     font-family: inherit;
     font-size: inherit;
     margin-bottom: 0;
+    text-decoration: none;
 }
 
 .btn {
     @extend %button-base;
-
-    text-decoration: none;
 }
 
-.btn--light {
-    @extend .btn-outline-light;
+.btn.btn--primary {
+    @extend .btn-outline-primary;
 
-    border: 1px solid $border-color;
-    background-color: $body-bg;
-    color: contrast-color($body-bg);
+    color: $text-color;
 
     &:focus,
     &:hover,
     &:active,
     &:focus-within,
     &:focus-visible {
-        background-color: $bg-tertiary !important;
+        background-color: $brand-primary-tint;
+        color: $text-color;
     }
 
     &:focus-visible {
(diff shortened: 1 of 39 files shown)


phillimorland:
This is another interim step as will have meeting with Marcelo tomorrow to discuss which styles should be kept and which should be done away with, still not sure if extending bootstrap btns was way to go rather then re-writting the old mixin but guess time will tell :(

apps/documents/assets/ChapterNavItem.jsx
```diff
@@ -17,7 +16,7 @@ const ChapterNavItem = (props) => {
     <div className="row" data-testid="chapter-nav-item">
       <button
         type="button"
-        className={classNames('commenting__content', 'commenting--toc__button', 'btn btn--light', 'btn--small', { active: props.active })}
+        className={(props.active && 'active') + 'commenting__content commenting--toc__button btn btn--light btn--small'}
```
fuzzylogic2000: is it fine like this or do we need a space between `active` and `commenting content`?

fuzzylogic2000: But cool, nice removal of package!

adhocracy-plus/assets/scss/components/_button.scss
```diff
@@ -4,103 +4,136 @@
...
 
 .btn {
     @extend %button-base;
-
-    text-decoration: none;
 }
 
-.btn--light {
-    @extend .btn-outline-light;
+.btn.btn--primary {
```
fuzzylogic2000, line 14: Why do these all have the btn and the variant? To make it more specific and overwrite other styles?

fuzzylogic2000 (commented):
Somehow, there was a diff outdated and I couldn't add a comment, so here some questions already and then I'll go on reviewing!

adhocracy-plus/assets/scss/components/_button.scss
phillimorland: @fuzzylogic2000 it was to make it more specific so we could still specify styling for just the `btn`class but it actually didn't work as planned, and i'm not sure if we should have anything more then basic styling assigned to btn anyway and as that was the black btn and given that it is difficult to show different states i'm not sure what to do for that one, once this is merged i'll let Marcelo know and we are going to have a chat about it tomorrow

phillimorland:
> Somehow, there was a diff outdated and I couldn't add a comment, so here some questions already and then I'll go on reviewing!

sorry no one was assigned so thought it wasn't being reviewed and pushed some updates

adhocracy-plus/assets/scss/components/_button.scss
```diff
@@ -4,103 +4,136 @@
...
     &:hover,
     &:active,
-    &:disabled,
-    &.is-disabled {
-        box-shadow: none;
-        background-color: $gray-lightest;
+    &:focus-within,
+    &:focus-visible {
+        background-color: $brand-tertiary !important;
+        color: invert($print-black) !important;
```
fuzzylogic2000, line 72: Why `invert($print-black)` instead of using some variable for white?

fuzzylogic2000 (approved):
Cool, many improvements! And one more question... :)

fuzzylogic2000:
> > Somehow, there was a diff outdated and I couldn't add a comment, so here some questions already and then I'll go on reviewing!
> 
> sorry no one was assigned so thought it wasn't being reviewed and pushed some updates

Right, sorry!

apps/documents/assets/ChapterNavItem.jsx
fuzzylogic2000: yes, found it! :)

adhocracy-plus/assets/scss/components/_button.scss
fuzzylogic2000: Cool, let's merge then!

PR #2108, opened 2022-11-29
WIP: [6751] templates/projects/module tile: redo staling of module tiles

fuzzylogic2000:


phillimorland:
@fuzzylogic2000 so I updated the icon content codes so they are displaying, not sure if they are correct though as didn't get all the codes? also it seems like there is the wrong number of them and possible one of them didn't work, I can't tell, also there seems a lot of crap in the svg file but not sure why?!

apps/users/templates/a4_candy_users/user_detail.html
```diff
@@ -110,7 +110,7 @@ <h2 class="u-heading--sm text-uppercase">{% trans 'Active in these platforms' %}
                       aria-expanded="true">
                       <ul class="l-tiles-2">
                       {% for project in view.projects %}
-                              {% include 'a4_candy_projects/includes/project_list_tile.html' with project=project orientation='vertical' %}
+                              {% include 'a4_candy_projects/includes/project_list_tile.html' with project=project type='project' object=project orientation='vertical' %}
```
fuzzylogic2000, line 113: @phillimorland good catch!

fuzzylogic2000 (commented):
Looks good to me! But someone else needs to review the whole thingincluding my commits :) 
@khamui @Rineee @goapunk

fuzzylogic2000:
Alright, @phillimorland and me decided to merge as we did review each other's commits. :)

PR #2111, opened 2022-11-30
webpack.common: update to maplibre -test

phillimorland:


phillimorland:
@goapunk test to see if libra works with mapbox as pre v 2 is meant to but it seems seems cors stops tile fetch:
`Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at mapbox://styles/liqd/ck3r7tmor14ux1cp07ne8gaaz. (Reason: CORS request not http).

Error: NetworkError when attempting to fetch resource.`

PR #2116, opened 2022-12-01
[updates/#6749] organisations/react_language_choice.jsx: fix linting error

khamui:


phillimorland (changes requested):
refactor looks good however the issue is still present, textarea not updating with active tabs, when you delete a language the textarea should not remain on the deleted language it should show the same textarea as the active tab, this definitely used to work although if i'm honest im not sure how long ago not sure if it was this update that broke it or another also since react update there is a nasty looking console error that might be connected which is still there in this PR

Issue #2013, opened 2022-08-01
weird line break


lthueer:
**URL:** https://adhocracy.plus/liqd/dashboard/projects/poll-for-office-renovation-feedback/basic/
**user:** 
**expected behaviour:** simple line break
**behaviour:** line break looks too big
**important screensize:**
**device & browser:** 
**Comment/Question:** 

![Bildschirmfoto 2022-08-01 um 14 10 58](https://user-images.githubusercontent.com/36957474/182144996-500fab46-318c-4c53-adf2-0d7e641ed232.png)

lthueer:
look also at #1980

PR #2125, opened 2022-12-06
css/_form.scss: removing spacing (margin-bottom) from input-base & ad…

khamui:
…d spacing to particular labels

closing #2013

### DIFF
--- a/adhocracy-plus/assets/scss/_form.scss
+++ b/adhocracy-plus/assets/scss/_form.scss
@@ -4,7 +4,6 @@ label {
 }
 
 %input-base {
-    margin: 0 0 $spacer;
     border: 1px solid $input-border-color;
     padding: 0.4em 0.6em;
     line-height: 1.6;


adhocracy-plus/assets/scss/_form.scss
```diff
@@ -4,7 +4,6 @@ label {
 }
 
 %input-base {
-    margin: 0 0 $spacer;
```
phillimorland, line 7: this will affect all inputs, do we want to do that? did you check different forms to make sure it doesn't break the styling everywhere?

phillimorland (changes requested):

adhocracy-plus/assets/scss/_form.scss
khamui: i double checked it in our components library. And it seems that it does not have any real effect on other inputs.

There is a class overwriting this to set it to 0 anyways. Except for type radio or checkbox. But those are not visible anyways due to own styling.

Bildschirmfoto 2022-12-08 um 16 41 06

PR #2131, opened 2022-12-06
dashboard/phases/datetime_input: using same class to assure same heig…

khamui:
…ht is applied to time input


phillimorland (approved):
coolio! just also added back a style i accidentally left commented out

PR #2134, opened 2022-12-08
project/tiles: tile tiles do not exceed width

khamui:


phillimorland (changes requested):
is there a reason we can't use `break-word` like we do in rest of code?

khamui:
> is there a reason we can't use `break-word` like we do in rest of code?

I tried `break-word` before, but somehow it has no effect here.

goapunk (approved):
works for me and there seems to be a lot of confusion about what the right thing to use is when you search for it. From what I gathered `anywhere` is fine.

PR #2135, opened 2022-12-08
[#6754] add img format and logo checkbox

phillimorland:


apps/organisations/templates/a4_candy_organisations/communication_form_social_media.html
```diff
@@ -24,8 +24,21 @@ <h1>{% translate "Create Content" %}</h1>
...
+                    <input class="checkbox-btn__input" type="checkbox" id="logo_1" value="add a+ logo" checked>
+                    <span class="checkbox-btn__text--icon">{% translate 'Add a+ logo' %}</span>
+                </label>
+            </div>
+            <div class="checkbox-btn">
+                <label class="checkbox-btn__label--light" for="logo_2">
+                    <input class="checkbox-btn__input" type="checkbox" id="logo_2" value="add own logo" checked>
+                    <span class="checkbox-btn__text checkbox-btn__text--icon">{% translate 'Add logo' %}</span>
+                </label>
+            </div>
```
phillimorland, line 38: @Rineee should I add these as a form_field include? I don't really know?!

apps/organisations/views.py
```diff
@@ -220,38 +220,49 @@ def form_valid(self, form):
 
     def generate_image(self, data):
         # Dummy images
-        image = Image.new('RGB', (304, 192), color='red')
+        image = Image.new('RGB', (1080, 760), color='red')
         image.save(settings.MEDIA_ROOT + '/images/tmp_image.png')
-        logo1 = Image.new('RGB', (100, 10), color='blue')
-        logo2 = Image.new('RGB', (100, 10), color='green')
+        logo1 = Image.new('RGB', (144, 144), color='blue')
+        logo2 = Image.new('RGB', (228, 56), color='green')
```
phillimorland, line 226: @Rineee i'm not sure how to get the logos so just used the dummy ones

apps/organisations/templates/a4_candy_organisations/communication_form_social_media.html
Rineee: No, I think it is fine like this, cool! 
Should we maybe grey out the 'add logo' if there is no organisation logo?

apps/organisations/templates/a4_candy_organisations/communication_form_social_media.html
```diff
@@ -43,19 +64,16 @@ <h1>{% translate "Create Content" %}</h1>
...
                 </div>
+                {% if image_preview %}
 
                 <div class="d-flex justify-content-end mb-5">
                     <button type="submit" class="btn btn--primary btn--icon">
-                        {% if image_preview %}
-                            <i class="fas fa-sync-alt"></i>{% translate 'Refresh' %}
-                        {% else %}
-                            {% translate 'Generate' %}
-                        {% endif %}
```
Rineee, line 53: Did you remove this on purpose @phillimorland ?

phillimorland: @Rineee  yeah because otherwise there ware 2 generate buttons which didn't make sense right? unless I accidentally added another?!

phillimorland: could that be what broke it? sorry when we talked about didn't mess with validation thought it was in the forms.py

apps/organisations/templates/a4_candy_organisations/communication_form_social_media.html
phillimorland: yes, good plan, you said that before but I forgot

apps/organisations/templates/a4_candy_organisations/communication_form_social_media.html
Rineee: Noo, dont think thats not what broke it!
I meant to add a 'Generate' button when there is no preview yet and a 'Refresh' button, when there is an preview, think it also works on dev? https://aplus-dev.liqd.net/liqd-tests/dashboard/communication/content/create/multimodul-test/format/1/

phillimorland: @Rineee ok I must have added another when correcting checkbox button position then messed up when trying to 'fix' it :p will undo and maybe squash all commit to clear it up a bit after I have the last configs from Marcelo

apps/organisations/views.py
```diff
@@ -215,44 +218,76 @@ def get_context_data(self, **kwargs):
...
+        # Get organisations logo
+        logo_org_get = Image.open(self.organisation.logo)
+        logo_org_size = (144, 144)
+        logo_org = logo_org_get.resize(logo_org_size)
+
+        # Get aplus logo
+        logo_aplus_get = Image.open(
+            os.path.join(
+                settings.STATIC_ROOT,
+                'images/logo.png'))
```
Rineee, line 249: This doesnt work for me as image is in assets instead of static?

phillimorland: @Rineee sorry you need to run `python manage.py collectstatic` or probably `make release` does it too

apps/organisations/views.py
```diff
@@ -215,44 +218,76 @@ def get_context_data(self, **kwargs):
 
     def form_valid(self, form):
         data = form.cleaned_data
-        img = self.generate_image(data)
         context = self.get_context_data()
+        img = self.generate_image(data)
+        # get the content form with unchanged cleaned data, as image generation
+        # changes the image data (why?)
+        content_form = CommunicationContentCreationForm(
+            initial=data, project=self.project, format=self.format)
```
phillimorland, line 226: @Rineee I still get weird error?! I will add all the correct configs and then look at this more

apps/organisations/views.py
Rineee: ah ups, sorry!

apps/organisations/views.py
Rineee: @phillimorland Hm, I checked again and for me it still works.. did you check with only these commits?

apps/organisations/forms.py
```diff
@@ -35,22 +35,38 @@
...
         'img_min_width': 1080,
-        'img_min_height': 1920},
-    3: {'title_max_length': 128,
-        'description_max_length': 256,
-        'img_min_width': 1104,
-        'img_min_height': 736},
-    4: {'title_max_length': 128,
-        'description_max_length': 256,
+        'img_min_height': 760,
+        },
```
phillimorland, line 51: there isn't configs for this one but may double check w marcelo this afternoon

apps/organisations/views.py
goapunk: we could also change the path to assets here so you don't have to run collectstatic to make your dev environment work (or do we need to do this anyway for other parts?)

apps/organisations/views.py
```diff
@@ -215,44 +218,93 @@ def get_context_data(self, **kwargs):
...
+            fill=(0, 0, 0),
+            font=fontsm)
+
+        if data['add_orga_logo']:
+            # Get organisations logo
+            logo_org_get = Image.open(self.organisation.logo)
+            logo_org_size = (144, 144)
+            logo_org = logo_org_get.resize(logo_org_size)
+            # position org logo
+            result.paste(logo_org, (160, 160))
```
goapunk, line 287: in the design there's a white border around the organization logo, this is not implemented yet, right?

phillimorland: @goapunk no I missed that from design, but this is already pretty big, i'll add it to a second PR

apps/organisations/forms.py
```diff
@@ -35,22 +35,46 @@
...
-        'description_max_length': 256,
-        'img_min_width': 1104,
-        'img_min_height': 736},
-    4: {'title_max_length': 128,
-        'description_max_length': 256,
+        'img_min_height': 1278,
+        'aplus_logo_width': 360,
+        'aplus_logo_height': 88,
+        'overall_height': 1920
+        },
```
phillimorland, line 57: NOTE: update styling of story to match new design and change name to 'Instagram story'

apps/organisations/views.py
```diff
@@ -215,44 +218,93 @@ def get_context_data(self, **kwargs):
...
         left = 0
-        top = 64
-        bottom = 48
+        top = 0
+        bottom = (
+            sharepic_format['overall_height'] -
+            sharepic_format['img_min_height'])
         width, height = image.size
         new_width = width + right + left
         new_height = height + top + bottom
```
goapunk, line 248: We don't need the calculations anymore I think, `width` will always be `img_min_width`  and `height` always `overall_height`.

goapunk, line 248: We don't need the calculations anymore I think, `width` will always be `img_min_width`  and `height` always `overall_height`.

apps/organisations/views.py
```diff
@@ -215,44 +218,93 @@ def get_context_data(self, **kwargs):
...
-        logo1 = Image.new('RGB', (100, 10), color='blue')
-        logo2 = Image.new('RGB', (100, 10), color='green')
-
-        # Adding padding
+        # Get required image size and image
+        sharepic_format = SOCIAL_MEDIA_SIZES[self.format]
+        image_get = Image.open(data['image'].file)
+        image_size = (sharepic_format['img_min_width'],
+                      sharepic_format['img_min_height'])
+        image = image_get.resize(image_size)
```
goapunk, line 237: we might want to use crop here to prevent weird stretching effects

apps/organisations/views.py
```diff
@@ -215,44 +218,93 @@ def get_context_data(self, **kwargs):
...
+            # Get organisations logo
+            logo_org_get = Image.open(self.organisation.logo)
+            logo_org_size = (144, 144)
+            logo_org = logo_org_get.resize(logo_org_size)
+            # position org logo
+            result.paste(logo_org, (160, 160))
+
+        if data['add_aplus_logo']:
+            # Get aplus logo
+            logo_aplus_get = Image.open(
```
goapunk, line 291: i think we should maybe load this from the assets folder, otherwise a fresh dev environment won't work out of the box with our Makefile.
One way could be:
```
logo_aplus_get = Image.open(
                os.path.join(
                    settings.BASE_DIR,
                    'adhocracy-plus/assets/images/logo.png'))
```
What do you think @Rineee @fuzzylogic2000

apps/organisations/views.py
```diff
@@ -215,44 +218,93 @@ def get_context_data(self, **kwargs):
...
-        # assigned to d1
-        d1 = ImageDraw.Draw(result)
-        # Text location, color and font
+        # image is converted into editable form using Draw function
+        image = ImageDraw.Draw(result)
+
+        # text dimension, color and font
         font = ImageFont.truetype(
-            "adhocracy-plus/assets/fonts/SourceSansPro-Semibold.otf", 20)
+            "adhocracy-plus/assets/fonts/SourceSansPro-Semibold.otf",
```
goapunk, line 258: we should use the same way as proposed for the a+ logo to get the path (or use this way for the logo), just for consistency

apps/organisations/views.py
goapunk, line 237: If the a+ logo is not included there is a lot of weird looking white padding at the bottom. One way to prevent this could be to increase the resize height here to something like `img_min_height` + `size of a+ logo`

goapunk:
another issue: if the a+ logo is not included there's too much white padding at the bottom, which makes it look weird. I'm not exactly sure how to solve this though

apps/organisations/views.py
Rineee: Yes, I think that would work!

phillimorland:
> another issue: if the a+ logo is not included there's too much white padding at the bottom, which makes it look weird. I'm not exactly sure how to solve this though

this is an issue for design, also it is preferable to have the logo so maybe it's a good thing

PR #2153, opened 2022-12-15
Pm 2022 12 styling issues

### LINKED ISSUE #2002: wagtail: columns cta block
**URL:** https://aplus-stage.liqd.net/
**user:** 
**expected behaviour:** when there is not a body text filled in the box is also not shown
**behaviour:** when columns cta is selected in wagtail for all CTA columns a body text is needed. If nothing is filled in the body. the little link box appears (see screenshot)
**important screensize:**
**device & browser:** 
**Comment/Question:*
This is only a problem for the newsletter that we link on the startpage. We select three columns and only use the one in the middle so that it's centered. Currently, we have found a workaround for prod by inserting a simple white image so that the body field is not empty.

Bildschirmfoto 2022-07-28 um 11 19 04

phillimorland:
fixes #2002

### DIFF
--- a/adhocracy-plus/assets/scss/components/_button.scss
+++ b/adhocracy-plus/assets/scss/components/_button.scss
@@ -37,7 +37,7 @@
 }
 
 %button-base {
-    padding: 0.4em 0.6em;
+    padding: (0.25 * $spacer) (0.75 * $spacer);
     line-height: 1.7;
     font-family: inherit;
     font-size: inherit;
--- a/adhocracy-plus/assets/scss/components/_button.scss
+++ b/adhocracy-plus/assets/scss/components/_button.scss
@@ -37,7 +37,6 @@
 }
 
 %button-base {
-    padding: (0.25 * $spacer) (0.75 * $spacer);
     line-height: 1.7;
     font-family: inherit;
     font-size: inherit;
--- a/adhocracy-plus/templates/a4modules/module_detail.html
+++ b/adhocracy-plus/templates/a4modules/module_detail.html
@@ -196,7 +196,7 @@
                         <ul class="l-tiles-3">
                             {% for other_module in module.module_cluster %}
                                 {% if not other_module == module %}
-                                    {% include "a4_candy_projects/includes/project_list_tile.html" with object=other_module project=other_module.project orientation='vertical' type='module' url=module.get_absolute_url %}
+                                    {% include "a4_candy_projects/includes/project_list_tile.html" with object=other_module project=other_module.project orientation='vertical' type='module' url=other_module.get_absolute_url %}
                                 {% endif %}
                             {% endfor %}
                         </ul>
--- a/apps/cms/pages/templates/a4_candy_cms_pages/blocks/col_cta_block.html
+++ b/apps/cms/pages/templates/a4_candy_cms_pages/blocks/col_cta_block.html
@@ -10,10 +10,11 @@
             <div class="block-col_cta-content">
                 <a href="{{ column.link }}" class="btn btn--full btn--transparent">{{ column.link_text }}</a>
             </div>
-            {% else %}
+            {% elif column.link %}
             <div class="block-col_cta-btn">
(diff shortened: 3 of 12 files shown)


adhocracy-plus/assets/scss/components/_button.scss
```diff
@@ -37,7 +37,7 @@
 }
 
 %button-base {
-    padding: 0.4em 0.6em;
+    padding: (0.25 * $spacer) (0.75 * $spacer);
```
khamui: cool, `()` improves readability

apps/cms/pages/templates/a4_candy_cms_pages/blocks/col_cta_block.html
```diff
@@ -10,10 +10,11 @@
             <div class="block-col_cta-content">
                 <a href="{{ column.link }}" class="btn btn--full btn--transparent">{{ column.link_text }}</a>
             </div>
-            {% else %}
+            {% elif not column.body and column.link %}
```
fuzzylogic2000, line 13: `{% elif column.link %}` is enough here, because the other one is taken care of with the ones above. (It will only get here if the conditions above are not True.)

adhocracy-plus/templates/a4modules/module_detail.html
```diff
@@ -196,7 +196,7 @@ <h3 class="tile__list-header">{%  translate 'More questions from this online par
                         <ul class="l-tiles-3">
                             {% for other_module in module.module_cluster %}
                                 {% if not other_module == module %}
-                                    {% include "a4_candy_projects/includes/project_list_tile.html" with object=other_module project=other_module.project orientation='vertical' type='module' url=module.get_absolute_url %}
+                                    {% include "a4_candy_projects/includes/project_list_tile.html" with object=other_module project=other_module.project orientation='vertical' type='module' url=other_module.get_absolute_url %}
```
fuzzylogic2000, line 199: oh, good catch!

apps/cms/pages/templates/a4_candy_cms_pages/blocks/col_cta_block.html
```diff
@@ -10,10 +10,11 @@
...
             <div class="block-col_cta-btn">
                 <a href="{{ column.link }}" class="btn btn--full btn--transparent">{{ column.link_text }}</a>
             </div>
+            {% else %}
```
khamui, line 17: does this case exist? `not column.body and not column.link`

apps/projects/templates/a4_candy_projects/project_list.html
```diff
@@ -11,7 +11,7 @@
         {% if filtered_projects|length > 0 %}
             <ul class="l-tiles-4">
                 {% for project in filtered_projects %}
-                    {% include "a4_candy_projects/includes/project_list_tile.html" with project=project %}
+                    {% include "a4_candy_projects/includes/project_list_tile.html" with object=project project=project orientation='vertical' type='project' url=project.get_absolute_url %}
```
khamui, line 14: is `object=project` needed?

apps/cms/pages/templates/a4_candy_cms_pages/blocks/col_cta_block.html
phillimorland: sorry forgot to add issue to commit https://github.com/liqd/adhocracy-plus/issues/2002

fuzzylogic2000 (approved):
Nice!
Only one nit-pick, which I am ok to ignore.

apps/projects/templates/a4_candy_projects/project_list.html
fuzzylogic2000: yes, we rewrote the tile include to use object. So that now they can be used for modules and projects.

phillimorland: @khamui I think so as the tile template uses `object.pk` and even if it's not useful in this instance, if we make changes to the tile template relying on list knowing the object we might not catch if it breaks something somewhere

fuzzylogic2000 (approved):
Nice! :upside_down_face:

PR #2163, opened 2022-12-19
Ks 2022 12 communication form tests

Rineee:
~~Setting this to WIP as it would be broken after image stuff is merged.. so will continue then.~~


fuzzylogic2000 (approved):
Very nice!

Issue #1870, opened 2022-05-30
a11y heading structure: project modules tiles


khamui:
**URL:** 
**user:** 
**expected behaviour:** heading hierarchy should be h1, h2, h3, h4 etc.
**behaviour:** heading h1 exists, but module-tile titles are h3, h2 is missing
**important screensize:**
**device & browser:** 
**Comment/Question:** 
Example see below "noinoi" is h1 and "Geschäftsord...." headings are h3. Missing h2 on this page.
Screenshot?
<img width="1654" alt="Bildschirmfoto 2022-05-30 um 13 13 35" src="https://user-images.githubusercontent.com/52801593/170990388-e30bd50d-d167-44b1-8f2d-b4e5c73ec728.png">

phillimorland:
It would make sense to have a title here, we could also add a sr only title, @mcastro-lqd design wise what do you think works here? should the tiles have a title explaining what they do?

mcastro-lqd:
@phillimorland @khamui I thought this would happen because of Multimodule projects, but its not the case (see screenshot), wouldn't it make more sense like that?  

Single: h1 - Main Project Title ➡️ h2 - Module Titles
Multimodule: h1 - Main project Title ➡️ h2 - Online Participation ➡️ h3 - Module Titles


<img width="1268" alt="Screenshot 2022-06-09 at 17 03 41" src="https://user-images.githubusercontent.com/59610786/172879611-eb4de626-c71f-4a60-874c-fd24e8966c3a.png">

phillimorland:
@mcastro-lqd just to confirm for multi-modules without the timeline (where modules all run at the same time) shall we add a h2 'Online Participation' heading?

mcastro-lqd:
@phillimorland I think it makes sense right? Since they can switch to a previous one

PR #2169, opened 2022-12-20
[Issues] templates/*: improve semantic html and add missing heading headings f…

phillimorland:
…ixes #1870 also fixes some a11y for the new tiles

### DIFF
--- a/apps/projects/templates/a4_candy_projects/includes/project_timeline-carousel.html
+++ b/apps/projects/templates/a4_candy_projects/includes/project_timeline-carousel.html
@@ -14,11 +14,11 @@
                         <div class="timeline-carousel__label {% if forloop.counter0 == initial_slide %} initial {% endif %}">
                             <span id="timeline_entry_{{ forloop.counter }}">
                                 {% if  date.type == 'module' %}
-                                    <span class="timeline-carousel__title">{{ date.title }}</span><br>
-                                    {% trans 'from' %} {{ date.date | date }}
+                                    <h2 class="timeline-carousel__title">{{ date.title }}</h2>
+                                    <date>{% trans 'from' %} {{ date.date | date }}</date>
                                 {% else %}
-                                    <span class="timeline-carousel__title">{{ date.event_type }}</span><br>
-                                    {% trans 'on' %} {{ date.date | date }}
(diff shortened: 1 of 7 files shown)


phillimorland:
@fuzzylogic2000 there's some weird stuff going on with the abbreviated times but when looking at it I didn't understand the different indexes when referencing the same time?! maybe we can look at it together or you can add a comment explanation?

apps/projects/templates/a4_candy_projects/includes/project_timeline-carousel.html
```diff
@@ -14,11 +14,11 @@
                         <div class="timeline-carousel__label {% if forloop.counter0 == initial_slide %} initial {% endif %}">
                             <span id="timeline_entry_{{ forloop.counter }}">
                                 {% if  date.type == 'module' %}
-                                    <span class="timeline-carousel__title">{{ date.title }}</span><br>
-                                    {% trans 'from' %} {{ date.date | date }}
+                                    <h2 class="timeline-carousel__title">{{ date.title }}</h2>
+                                    <date>{% trans 'from' %} {{ date.date | date }}</date>
```
fuzzylogic2000, line 18: Not new, but seeing this: we should probably use blocktranslate here.

apps/projects/templates/a4_candy_projects/project_detail.html
```diff
@@ -161,6 +161,8 @@ <h3 class="u-top-divider">{% trans 'Contact for questions' %}</h3>
                 <div class="timeline-carousel__wrapper">
                 {% if project.display_timeline %}
                     {% include "a4_candy_projects/includes/project_timeline-carousel.html" with dates=project.participation_dates project=project %}
+                {% else %}
+                    <h2 class="visually-hidden">{% translate 'Online articipation' %}</h2>
```
fuzzylogic2000, line 165: There is a p missing here 'Online participation'

fuzzylogic2000 (changes requested):
Looks good! Only one typo that needs to be fixed first.
I will add an explanation for the ebbreviated stuff and the blocktranslate in an extra branch.

fuzzylogic2000 (approved):
Noice!

PR #2179, opened 2022-12-22
sharepics: add test for aspect ratio calculation

goapunk:


apps/organisations/views.py
```diff
@@ -283,13 +289,11 @@ def generate_image(self, data):
 
         # add text using width to center
         image.text(((sharepic_format['img_min_width'] - title_width) / 2,
-                    sharepic_format['title_y']),
-                   data['title'], fill=(0, 0, 0), font=font)
+                    sharepic_format['title_y']), title, fill=(0, 0, 0),
+                   font=font)
```
fuzzylogic2000, line 293: The brackets are a bit hard to read. Can we use the same format for this one and the one below? I personally like:
```
some_function(
    stuff_here,
    (stuff in more braces, bla),
    and_so_on
)
```
That will not quite work here because of the longer brace in the brace, but at least have everything eles in single lines?

fuzzylogic2000 (commented):
Cool! Only one nit-pick and also maybe rebase and check if still not failing after I merged @phillimorland redoing of the sizes?

apps/organisations/views.py
```diff
@@ -282,13 +288,16 @@ def generate_image(self, data):
...
         # add text using width to center
-        image.text(((sharepic_format['img_min_width'] - title_width) / 2,
-                    sharepic_format['title_y']),
-                   data['title'], fill=(0, 0, 0), font=font)
+        image.text(
+            ((sharepic_format['img_min_width'] - title_width) / 2,
+                sharepic_format['title_y']),
+            title,
+            fill=(0, 0, 0),
+            font=font)
```
goapunk, line 296: still not pretty but maybe a bit better?

phillimorland (approved):
looking good! :)

apps/organisations/views.py
fuzzylogic2000: Yes, that's easier to read!

fuzzylogic2000 (approved):
Nice!

PR #2183, opened 2022-12-22
templates/buttons: using <button>s instead of <input>s as click eleme…

khamui:
…nts (fixes no. 7 of issue 2141 appearance on mobile devices)


fuzzylogic2000 (approved):
Nice! Should we also do that in mB?

PR #2184, opened 2022-12-22
Pm 2022 12 story insta

phillimorland:


apps/organisations/templates/a4_candy_organisations/organisation_form.html
```diff
@@ -5,7 +5,7 @@
 
 {% block organisation_form %}
     <div class="col-md-10">
-        <h1>{% trans "Edit your organisation information" %}</h1>
+        <h1 class="u-first-heading">{% trans "Edit your organisation information" %}</h1>
```
goapunk, line 8: when touching lines like this we could also do trans -> translate :)

goapunk (approved):
nice, looks much better

PR #2185, opened 2022-12-22
replace trans with translate

goapunk:


goapunk:
see https://taiga.liqd.net/project/magdan-technical-backlog/us/61?no-milestone=1

fuzzylogic2000 (approved):
Nice! `makemessages` doesn't loose anything, so it must be fine.

PR #2195, opened 2022-12-30
Ks 2022 12 more issues

Rineee:
The third commit fixes  #1073 as this error happens, because the subjects dont get annotated (the annotation happens in the filter and if the filter argument is invalid, it doesnt get there). Then, when adding the annotations, I thought, I could also move the filtering for module to the parent class, but that actually broke the .count in the template :woman_shrugging:  Now it uses for/empty tags which works!


fuzzylogic2000 (approved):
Supercool!

Rineee:
@fuzzylogic2000 Now it should be fine again!

PR #2202, opened 2023-01-03
WIP: replace datepicker with flatpickr

goapunk:
replace datepicker with flatpickr as it's actively maintained


goapunk:
closing as outdated

PR #2203, opened 2023-01-03
[6876] remove django-capture-tag

fuzzylogic2000:
For the email content, I did change a few translations and removed the newlines and html tags from the strings where appropriate.
And I noticed that we forget the `blocktranslate`, so I also changed that.


khamui:
Very nice! LGTM

PR #2209, opened 2023-01-05
apps/embed: remove embed code

fuzzylogic2000:
I hope, I found everything! But I have questions (added that to the comments, so the lines that are new are the questions). Could you have a look @phillimorland, @khamui  or @goapunk ?


adhocracy-plus/assets/js/popup-close.js
```diff
@@ -1,4 +1,5 @@
 // if this page was opened from an embed for login, notify it about success
+// does this mean we can remove the whole function now?
```
khamui, line 2: yes, think this file can go

khamui (approved):
looking good to me! but will let someone with more proficiency about <kbd>embed</kbd> have a look :)

Question:
Should we leave the `docs/embed.md` and make a note there, that this feature has been removed? So in case someone is wondering where the embed went, it is at least documented there?
Hmm..but maybe this will be in the release notes anyways.

adhocracy-plus/assets/js/unload_warning.js
```diff
@@ -26,6 +26,7 @@ document.addEventListener('DOMContentLoaded', () => {
   }
 
   // This clashes with the embed plugin and so is commented out for now until we have time to fix
+  // It's not commented out??
```
phillimorland, line 29: oooo so @goapunk un-commented it and I merged it but not sure how neither of us caught that, I think it was working after the vanilla refactor, does that sound right @goapunk?

phillimorland (approved):
awesome! I think there maybe some bits to remove in a4 too but guess we have to wait on mB there, can be merged after the file is deleted

adhocracy-plus/assets/js/unload_warning.js
goapunk: yes, I think we left it like this because it was working? or because we knew the embed was about to be removed ? :shrug:

fuzzylogic2000:
Let's merge after the release...

fuzzylogic2000:
@phillimorland @khamui I did a rebase. Could you have another look before merge?

phillimorland (approved):
Yey!

PR #2218, opened 2023-01-10
Ks 2023 01 add black

Rineee:
Soo, this is the code formatted by black.. What do you think?

I actually had to commit now with --no-verify, because black does not always meet the 88 char rule.. I am unsure what to do about it,
1. allow to be over 88 chars sometimes (in flake8)
2. manually refactor to always meet 88 chars.. (this is what was done in django project)

I think I would even prefer option 1, as it isnt that many times it happened (I think 7 times in the whole formatted code)

Oh, and in case we merge this, there is an [option to not ruin git blame](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame), which I will then do afterwards:


apps/ideas/views.py
```diff
@@ -79,22 +71,23 @@ class IdeaListView(AbstractIdeaListView, DisplayProjectOrModuleMixin):
...
 
 
 class IdeaDetailView(AbstractIdeaDetailView):
     model = models.Idea
-    queryset = models.Idea.objects.annotate_positive_rating_count()\
-        .annotate_negative_rating_count()
-    permission_required = 'a4_candy_ideas.view_idea'
+    queryset = (
+        models.Idea.objects.annotate_positive_rating_count().annotate_negative_rating_count()
+    )
```
Rineee, line 84: This line is for example longer than 88, but tbh I dont know how to change that (just moving a part of it to the next line is still considered to be one line in flake), unless writing a function annotate_rating_counts() :) Do you have an idea?

fuzzylogic2000: maybe 
```
queryset = models.Idea.objects.all()
queryset.annotate_positive_rating_count()
queryset.annotate_negative_rating_count()
```

fuzzylogic2000 (approved):
It doesn't make everything easier to read, but everything will have the same formatting afterwards, which is good. And I think it will save us time in the future.

fuzzylogic2000:
but let's merge after the release. :)

apps/ideas/views.py
Rineee: That doesnt work as it is just the defintion of the class attributes.. I can maybe try and put it into innit? Will do!

Rineee:
As discussed with @fuzzylogic2000 I now added the ignore=E501 to flake8 config, so we now allow lines longer than 88 chars.

fuzzylogic2000 (approved):
Nice!

PR #2225, opened 2023-01-12
translations: fix missing strings

### LINKED ISSUE #2221: stage: translations missing
**URL:** https://aplus-stage.liqd.net/testorga-luca/dashboard/communication/content/create/sadf-3/format/1/
**user:** 
**expected behaviour:** all translations are there
**behaviour:** translations are missing
**important screensize:**
**device & browser:** 
**Comment/Question:** 

- dev & stage
- translations for module type are there in dashboard (only missing on tile)
- translation strings are on Transifex but not on a+

helptexts fpr sharepics
Bildschirm­foto 2023-01-11 um 15 28 38

type of module not translated
Bildschirm­foto 2023-01-11 um 15 28 11

fuzzylogic2000:
Fixes #2221

First commit:
I am pretty unhappy with this, but I didn't find way to make it work with the interpolated strings. I tried using `gettext` instead of `gettext_lazy` and changing the interpolation method to the named one with %, but both didn't work. Any other ideas @Rineee ?
Edit:
I discussed in the issue with Luca. It would be fine doing it like that. I am still not overly happy, but guess it's alright.


Rineee:
not needed anymore because of https://github.com/liqd/adhocracy-plus/pull/2230

PR #2229, opened 2023-01-13
Revert "add git-blame-ignore-revs file"

fuzzylogic2000:
Reverts liqd/adhocracy-plus#2228


Rineee:
Yeah, probably best to revert for now..

PR #2230, opened 2023-01-13
organisations/forms: move brackets to right place to make string tran…

Rineee:
…slatable

fixes https://github.com/liqd/adhocracy-plus/issues/2221
closes https://github.com/liqd/adhocracy-plus/pull/2225


apps/organisations/forms.py
```diff
@@ -314,10 +314,8 @@ def __init__(self, project=None, format=None, *args, **kwargs):
             required=True,
             help_text=_(
                 "This title will be displayed as a header. "
-                "It should be max. {} characters long.".format(
-                    self.sizes["title_max_length"]
-                )
-            ),
+                "It should be max. {} characters long."
```
fuzzylogic2000, line 317: Ah!

locale-source/locale/de/LC_MESSAGES/django.po
```diff
@@ -893,24 +893,24 @@ msgid "seconds"
...
 #: adhocracy4/modules/models.py:227
-msgctxt "abbreviation for hour(s)"
+msgctxt "abbreviation \"hour(s)\""
 msgid "H"
-msgstr "S"
+msgstr ""
 
 #: adhocracy4/modules/models.py:229
-msgctxt "abbreviation for minute(s)"
+msgctxt "abbreviation \"minute(s)\""
```
fuzzylogic2000, line 906: This is weird, when I do an install and then do the "makemesages", these explanations are not changed and the translations not lost.

fuzzylogic2000 (approved):
Supercool!
I will redo the makemessages and then also push to transifex!

locale-source/locale/de/LC_MESSAGES/django.po
Rineee: Oh, the translations were also lost? Thats not good..

PR #2236, opened 2023-01-16
settings/base: change wording of a4_blueprint_types

### LINKED ISSUE #2235: Transifex module Tiling translations differently
**wording on transifex for multi modules:**

- brainstorming
- map brainstorming
- idea collection
- map idea collection
- text review
- poll
- interactive event
- topic prioritization
- debate

**wording how it should be (in line with dashboard):**

- Brainstorming
- Spatial Brainstorming
- Idea Challenge
- Spatial Idea Challenge
- Text Review
- Poll
- Interactive Event
- Prioritization
- Debate

Participatory Budgeting is fine and already translated correctly

Rineee:
@fuzzylogic2000 Could you add the makemessages again? It is still doing the weird thing with the abbreviations for me :woman_shrugging: Also, do we want to add the strings uppercase here already? Wasnt sure..

fixes  #2235

### DIFF
--- a/adhocracy-plus/config/settings/base.py
+++ b/adhocracy-plus/config/settings/base.py
@@ -515,14 +515,14 @@ A4_CATEGORY_ICONS = (
 
 A4_BLUEPRINT_TYPES = [
     ("BS", _("brainstorming")),
-    ("MBS", _("map brainstorming")),
-    ("IC", _("idea collection")),
-    ("MIC", _("map idea collection")),
+    ("MBS", _("spatial brainstorming")),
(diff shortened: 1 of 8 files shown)


fuzzylogic2000:
> @fuzzylogic2000 Could you add the makemessages again? It is still doing the weird thing with the abbreviations for me woman_shrugging Also, do we want to add the strings uppercase here already? Wasnt sure..
> 
> fixes #2235

Hmm, don't know about the uppercases. Maybe? But I think this is good as well! Let's keep them small.

PR #2250, opened 2023-01-24
[#6951]scss/modal: styling modal to match design (depending on a4)

khamui:


apps/contrib/templates/a4_candy_contrib/component_library.html
```diff
@@ -1762,7 +1781,7 @@ <h2 id="cl-multiform">Multiform</h2>
                               <input type="email" name="email" placeholder="E-mail address" class="form-control" required="" id="id_email">
                           </div>
                           <div class="d-flex justify-content-end mb-3">
-                            <button class="btn btn--primary" name="action_add" type="submit">{% trans 'Add Email' %}</button>
```
khamui, line 1765: known issue!

khamui:
Can be reviewed, but please do not merge before next Release.

fuzzylogic2000:
What is the status here? It's for the a+ part for the stuff we changed in a4? Did we do the same in mB? Maybe who did the reviewing there can also review this? I hope it was @phillimorland or @Rineee ?

Rineee:
> What is the status here? It's for the a+ part for the stuff we changed in a4? Did we do the same in mB? Maybe who did the reviewing there can also review this? I hope it was @phillimorland or @Rineee ?

I think it was @phillimorland :see_no_evil:

adhocracy-plus/assets/scss/components/_modal.scss
```diff
@@ -0,0 +1,48 @@
...
+    button {
+        margin-top: 0;
+        margin-bottom: 0;
+    }
+
+    button.submit-button {
+        @extend .btn;
+        @include button-bg($print-black, $print-black, $body-bg, 15%);
+        margin-right: 0;
+        order: 2;
```
phillimorland, line 19: we already have styling for these buttons and we want them to be consistent so I would not add special styles for the modals, instead maybe move the existing selectors to buttons.scss and extend the button styles we will use, setting our own will mean we might get inconsistencies in our button styles again, this is something it would be great to clean up so we mostly just want to extend our existing styles

apps/dashboard/templates/a4_candy_dashboard/includes/progress.html
```diff
@@ -22,32 +22,40 @@
...
-                        </button>
-                    </form>
-
-                </div>
+            </div>
+            <div class="modal-footer">
+                <button type="button" data-bs-dismiss="modal" class="cancel-button">{% translate 'Cancel'%}
+                </button>
+                <form class="submit-form" action="{% url 'a4dashboard:project-publish' organisation_slug=project.organisation.slug project_slug=project.slug %}" method="post" data-ignore-submit="true" class="u-inline">
+                    {% csrf_token %}
```
phillimorland, line 53: this element has 2 classes so assume at lease 1 doesn't work

adhocracy-plus/assets/scss/components/_modal.scss
```diff
@@ -0,0 +1,48 @@
...
+            margin-right: 0;
+        }
+    }
+
+    button.cancel-button {
+        @extend .btn;
+        @include button($brand-tertiary, $text-color, $text-color, 57%);
+        margin-left: 0;
+        order: 1;
+    }
```
phillimorland, line 37: do we need the element and classes in this file?

phillimorland (commented):
so I think it can be a bit neater with button stuff being moved and maybe we can trim down the modal css. Also I have an issue with the order of the buttons on the react components but i'm not sure if thats my local setup?

adhocracy-plus/templates/a4dashboard/includes/project_list_item.html
```diff
@@ -79,17 +79,29 @@ <h3 class="u-first-heading">
...
-    <div class="modal-dialog" role="document">
+    <div
+        class="modal-dialog modal-dialog-centered"
+        role="document"
+    >
         <div class="modal-content">
             <div class="modal-header">
-                <h2 class="modal-title" id="project{{ project.pk }}DeleteModal">{% blocktranslate with project_name=project.name %} Are you sure you want to delete the project '{{ project_name }}'? {% endblocktranslate %}</h2>
-                <button type="button" class="btn btn--close" data-bs-dismiss="modal" aria-label="Close"></button>
+                <h2 className="u-no-margin-bottom u-spacer-top-one-half">
```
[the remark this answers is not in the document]
khamui: hm..yes, but that would affect mB as well right? also it would affect the Modal (and its siblings, e.g. ReportModal etc.) component, as I am using the same classes...

adhocracy-plus/assets/scss/components/_modal.scss
khamui: this modal specific classes are coming from mB and I think it makes more sense, because otherwise we would tie both projects to the same class names which could lead to problems for instance in mB `btn btn--primary` and here `btn btn--light`.

I think we are more flexible to extend these modal button classes with the specific ones needed in each project.

Here I just didn't manage to `extend from two classes`, that is not ideal I agree...Not sure how to extend it better, rather having copied the styles (in this case `@include button(....`)

adhocracy-plus/assets/scss/components/_modal.scss
khamui: will remove them and see if it still applies

apps/dashboard/templates/a4_candy_dashboard/includes/progress.html
khamui: oh, oops. good catch

adhocracy-plus/templates/a4dashboard/includes/project_list_item.html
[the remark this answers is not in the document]
phillimorland: it can be documented here without touching the other repos?!

adhocracy-plus/assets/scss/components/_modal.scss
phillimorland: submit-button in a bootstrap button I think but we don't want to have multiple different submit buttons, no matter where they are as they will hopefully do the same thing so we should use our standard classes btn and btn--default or whichever suits the design and as they are buttons they can be moved to button.scss and also we can not double up with them in shame.scss too

khamui:
wip - tests are failing

fuzzylogic2000 (approved):
very nice! 
I did rebase to try it out, so will push that and merge.

PR #2253, opened 2023-01-26
Kl 2023 01 release prep

fuzzylogic2000:


locale-source/locale/de/LC_MESSAGES/djangojs.po
```diff
@@ -807,23 +806,22 @@ msgid_plural "%s people have answered."
 msgstr[0] "1 Person hat geantwortet."
 msgstr[1] "%s Menschen haben geantwortet."
 
-#: adhocracy4/polls/static/PollDetail/PollResults.jsx:80
-#: adhocracy4/polls/static/PollDetail/PollResults.jsx:80
-msgid "noone has answered this question"
-msgstr "niemand hat diese Frage beantwortet"
+#: adhocracy4/polls/static/PollDetail/PollResults.jsx:78
+msgid "no one has answered this question"
```
fuzzylogic2000, line 810: Translations are driving me insane again! This is right, @phillimorland did fix that string in a4 and I wanted to add the old trsnalation to then get then on prod and into transifex.

fuzzylogic2000 (commented):
I will ignore that translation mixup and also add the new string, so that it will still be translated, but whyyyyyy? I guess we should still investigate. Any ideas @Rineee  @goapunk or @khamui ?

Well, update: with `make clean`, everything is now how it should be. So, for future reference: new a4 js strings do not only require an install, but also the deletion of the node_modules to be correct. :roll_eyes:

locale-source/locale/de/LC_MESSAGES/djangojs.po
fuzzylogic2000: Great, I also didn't add the second comment here. The old string was also still there. I will fix the whole commit now, though, so it's not anymore.
So, in short: all is fine now. But without a `make clean` and new install before the `make po`, old a4 strings are still in the djangojs.po files.

PR #2272, opened 2023-02-07
apps/*/dashboard: using blueprint types for export component

khamui:


fuzzylogic2000 (commented):
Oh, we didn't even use the types here yet... Cool!
But there seems to be sth. wrong with one of the dashboard components in the interactive event?!

apps/budgeting/dashboard.py
```diff
@@ -14,9 +14,8 @@ class ExportBudgetingComponent(DashboardComponent):
     label = _("Export Excel")
 
     def is_effective(self, module):
-        module_app = module.phases[0].content().app
         return (
-            module_app == "a4_candy_budgeting"
+            module.blueprint_type == "PB"
```
Rineee, line 18: dont we need all budgeting types here?

khamui:
> Oh, we didn't even use the types here yet... Cool!
> But there seems to be sth. wrong with one of the dashboard components in the interactive event?!

indeed. I don't get why it is failing if I am checking the `blueprint_type`. Any idea @fuzzylogic2000 @Rineee ?

apps/budgeting/dashboard.py
khamui: I think in aplus there is one phase PB available? at least in the Blueprint list I cannot find another one

fuzzylogic2000 (approved):
Cool! 
(You did look at them all in the dashboard and everything is where it should be, right? Then, please merge!)

khamui:
will doublecheck if everything is as it should be.
I don't get why the coverage is going down? it should go up instead right? because the absolute number of lines decreased...

fuzzylogic2000:
> will doublecheck if everything is as it should be. I don't get why the coverage is going down? it should go up instead right? because the absolute number of lines decreased...

But we have less tested lines, because we shortened the code and with that replaced two tested lines by one (in a few files). So the untested lines became a bigger part of all the code lines. :woman_shrugging:

PR #2283, opened 2023-03-06
assets/shame: make sure comment buttons are styled

fuzzylogic2000:
Decided to fix it here instead of adding classes in a4.


khamui:
cool!

PR #2312, opened 2023-03-29
Ks 2023 03 add userdashboard

Rineee:
Sooo, the first commit is really huge, but it was hard to cut the line, so I just added (almost) everything in there.. do you think that is ok?
Also there is a lot that still needs to be done, eg the renaming of stuff in all the jsx files. I only did it where absolutely necessary (mostly in ModerationNotification.jsx). Also, I did not yet remove the kosmo specific translations as that would have added even more files..and the styling is broken obviously.

### DIFF
--- a/adhocracy-plus/config/settings/base.py
+++ b/adhocracy-plus/config/settings/base.py
@@ -101,6 +101,7 @@ INSTALLED_APPS = (
     "apps.exports",
     "apps.offlineevents",
     "apps.projects",
+    "apps.userdashboard",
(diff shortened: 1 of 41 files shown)


fuzzylogic2000 (approved):
Yeah, it does work! And the comments are there.
I think we should merge and go on from there. :) Can I merge?

Rineee:
> Yeah, it does work! And the comments are there. I think we should merge and go on from there. :) Can I merge?

Yes please!

PR #2313, opened 2023-03-30
apps/users//views//models//templates: using profile view as in Kosmo

khamui:


fuzzylogic2000 (changes requested):
Nice, works perfectly! 
Only thing: the strings shouldn't be marked as kosmo-strings anymore. And a time-decision. :clock1:

apps/users/models.py
```diff
@@ -128,6 +130,37 @@ class User(auth_models.AbstractBaseUser, auth_models.PermissionsMixin):
     USERNAME_FIELD = "email"
     REQUIRED_FIELDS = ["username"]
 
+    def get_projects_follow_list(self, exclude_private_projects=False):
```
fuzzylogic2000, line 133: @Rineee also added that in the other PR, but git seems to be fine with that. :)

fuzzylogic2000 (approved):
Nice!

PR #2315, opened 2023-04-03
apps/users/templates: adding manual link (from settings) under profil…

khamui:
…e avatar and copy scss file


fuzzylogic2000 (changes requested):
It looks good. But I cannot see the button.

fuzzylogic2000 (commented):
I only notoced this time that you added the link in the user profile. It was meant to be in the userdashboard, like in kosmo. I changed that.

PR #2330, opened 2023-04-05
[7210] userdashboard: add filters for moderation dashboard in api and use in…

Rineee:
… frontend

Depends on https://github.com/liqd/adhocracy4/pull/1384

When testing the sorting, remember that the date displayed on the tile might be the modified date. I changed the serializer to always return created for testing.


fuzzylogic2000 (approved):
Supercool!

PR #2343, opened 2023-04-11
[#7208] React18 syntax and rm un-needed js from being loaded (a11y related)

philli-m:


apps/userdashboard/assets/js/a4_candy_userdashboard/react_moderation_notification_list.jsx
```diff
@@ -1,22 +1,45 @@
...
+        />)
+    })
 }
 
 document.addEventListener('DOMContentLoaded', init, false)
+
+// function init () {
+//   ReactWidget.initialise('aplus', 'moderation_notification_list',
+//     function (el) {
+//       const props = JSON.parse(el.getAttribute('data-attributes'))
```
philli-m, line 29: @goapunk @khamui this doesn't work and i'm not sure why, it's more a point of curiosity then anything but wanted to ask if there's anything i'm missing as it works in other apps, when this is used then props is empty, however it will work with `const props = JSON.parse(JSON.stringify(el.dataset))` not sure why but as dataset is meant for less complicated data structures I though I shouldn't introduce it, anyway will finish off removing excess js

adhocracy-plus/assets/js/app.js
```diff
@@ -31,11 +29,6 @@ function init () {
   ReactWidget.initialise('a4', 'ratings', ReactRatings.renderRatings)
   ReactWidget.initialise('a4', 'reports', ReactReports.renderReports)
```
philli-m, line 31: I think we should also remove Polls and possible reports from here too but as they come from a4 it's a little trickier, will add a pointless task

apps/interactiveevents/templates/a4_candy_interactive_events/livequestion_module_detail.html
```diff
@@ -36,3 +36,8 @@ <h3>{{ module.name }}</h3>
 </div>
 {% include 'footer_upper.html' with organisation=module.project.organisation %}
 {% endblock %}
+
+{% block extra_js %}
+    <script type="text/javascript" src="{% static 'interactiveevents.js' %}"></script>
+    {{ block.super }}
+{% endblock %}
```
goapunk, line 43: missing newline at the end here

apps/userdashboard/assets/js/a4_candy_userdashboard/react_moderation_notification_list.jsx
goapunk: I'm not sure I follow. The `moderation_notification_list` doesn't have a `data-attributes`, so it also doesn't find one here so I assume it's a typo :) what works is `const props = el.dataset`, no need to parse

philli-m: @goapunk not a typo just me being dumb, thought data-attributes should fetch all the data-*- attributes. Anyway will swap to using dataset then if are not worried about introducing new syntax, cheers!

apps/documents/assets/react_documents_init.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import { createRoot } from 'react-dom/client'
+import DocumentManagement from './DocumentManagement'
+import { widget as ReactWidget } from 'adhocracy4'
+
+function init () {
+  ReactWidget.initialise('mb', 'document-management',
+    function (el) {
+      const chapters = JSON.parse(el.getAttribute('data-chapters'))
```
goapunk, line 9: if we start using `e.dataset` I think we should use it everywhere ?

goapunk (approved):
cool

PR #2344, opened 2023-04-11
[#7208] styling: userdashboard: fix project list tiles and tab location, clean comp li…

philli-m:
…b and rm uneeded styles


adhocracy-plus/assets/scss/components/_filter.scss
```diff
@@ -0,0 +1,7 @@
+.filter--full {
+    margin-top: $spacer;
+
+    & button {
+        width: 100%;
+    }
+}
```
philli-m, line 7: I would like to move all filter styling to here eventually so starting now

adhocracy-plus/templates/a4dashboard/blueprint_list.html
```diff
@@ -1,11 +1,11 @@
-{% extends "a4dashboard/base_dashboard.html" %}
+<!-- {% extends "a4dashboard/base_dashboard.html" %}
```
khamui, line 1: is a closing comment missing here?

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationList.jsx
```diff
@@ -161,42 +161,52 @@ export default class ModerationNotificationList extends Component {
...
+        </div>
+
+        <div
+          className="row"
+          aria-label={django.pgettext('kosmo', 'Filter')}
+        >
+          <div className="col-md">
             <Filter
-              filterClass="mt-3 mt-lg-5 me-lg-3 dropdown dropdown-menu-end"
+              filterClass="filter--full dropdown dropdown-menu-end"
```
khamui, line 183: out of curiosity: is there a general rule we follow when deciding, when to use bs and when to use our own classes?

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
```diff
@@ -255,8 +257,8 @@ export const ModerationNotification = (props) => {
   }
 
   return (
-    <>
-      <li className="list-item">
+    <li className="list-item border-0 p-0">
+      <div className="list-item">
```
khamui, line 261: is this div still needed and should it use the same class `list-item` as its parent?

khamui (commented):
the three columns spacing looks a little bit off (spacing to bottom is too less? and to the right edge is too big?)

And the image looks a little bit wider than in the design (nit-picky, but I thought I mention it)

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotificationList.jsx
philli-m: @khamui that's a good question! the answer is kind of but not really, in this case the BS classes `dropdown dropdown-menu-end` are needed for the functionality so unless it is general styling that should be applied to every dropdown we should not add anything to that class so having an additional class is preferable. Also as in this case we will want all children of filter--full to be full width, we add a class as this is simpler then assigning bs classes in multiple places. I don't think it's particularly standard to have such a mix as we do and finding the right medium is pretty hard but as bootstrap can be very long winded to get something precise it's kind of if it 'feels' right then we go with bs but if not with a component class.

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx
philli-m: the div is needed to ensure that the mod statement and comment are in the same border box but the comment box for the statement is external as in the designs but you are right, double use of list-item is not great, i'll go back and reconsider

khamui:
lgtm!

PR #2347, opened 2023-04-12
userdashboard: improve db queries for actions

goapunk:
this should help at least a little with the very slow userdashboard

### DIFF
--- a/apps/projects/api.py
+++ b/apps/projects/api.py
@@ -4,10 +4,9 @@ from rest_framework import viewsets
 from adhocracy4.modules.models import Module
 from adhocracy4.projects.models import Project
 
-from . import helpers
 from .serializers import AppModuleSerializer
 from .serializers import AppProjectSerializer
-from .serializers import ProjectSerializer
+from .serializers import ModerationProjectSerializer
 
 
 class AppProjectsViewSet(viewsets.ReadOnlyModelViewSet):
@@ -30,12 +29,8 @@ class AppModuleViewSet(viewsets.ReadOnlyModelViewSet):
 
 
 class ModerationProjectsViewSet(viewsets.ReadOnlyModelViewSet):
-    serializer_class = ProjectSerializer
+    serializer_class = ModerationProjectSerializer
     permission_classes = (permissions.IsAuthenticated,)
 
     def get_queryset(self):
-        project_report_count = {
-            project: helpers.get_num_reports(project)
-            for project in list(self.request.user.project_moderator.all())
-        }
-        return sorted(project_report_count, key=project_report_count.get, reverse=True)
+        return self.request.user.project_moderator.all().select_related("organisation")
--- a/apps/projects/serializers.py
+++ b/apps/projects/serializers.py
@@ -194,7 +194,7 @@ class AppModuleSerializer(serializers.ModelSerializer):
         return False
 
 
-class ProjectSerializer(serializers.ModelSerializer):
+class ModerationProjectSerializer(serializers.ModelSerializer):
     title = serializers.SerializerMethodField()
     organisation = serializers.SerializerMethodField()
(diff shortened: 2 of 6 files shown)


apps/projects/api.py
```diff
@@ -30,12 +29,8 @@ def get_queryset(self):
 
 
 class ModerationProjectsViewSet(viewsets.ReadOnlyModelViewSet):
-    serializer_class = ProjectSerializer
+    serializer_class = ModeratorProjectSerializer
     permission_classes = (permissions.IsAuthenticated,)
 
     def get_queryset(self):
-        project_report_count = {
-            project: helpers.get_num_reports(project)
```
goapunk, line 38: from what I understand this doesn't actually work? the number of reports isn't sent as the serializer drops it. I moved it in the serializer and added the sorting to the front end

apps/userdashboard/views.py
```diff
@@ -45,9 +44,6 @@ def projects(self):
         projects = Project.objects.filter(
             follow__creator=self.request.user, follow__enabled=True
         )
-        for project in projects:
-            project.latest_comments_num = helpers.get_num_latest_comments(project)
```
goapunk, line 49: I couldn't find any place where this was used so I removed it. Or did I miss this somehow?

apps/projects/api.py
philli-m: it also seems to be in helpers and api? but that's all i know

apps/userdashboard/views.py
philli-m: I think we maybe used to use it to say so many comments in the last week but don't think we use it anymore but maybe it can also be removed from helpers.py

apps/projects/serializers.py
```diff
@@ -332,3 +332,30 @@ def get_moderation_detail_url(self, instance):
         return reverse(
             "userdashboard-moderation-detail", kwargs={"slug": instance.slug}
         )
+
+
+class ModeratorProjectSerializer(ProjectSerializer):
+    num_reports = serializers.SerializerMethodField()
```
Rineee, line 338: I think we only use the ProjectSerializer for the Moderation Dashboard, so you could just change that one. But renaming it is actually a good idea!

Rineee (commented):
The database optimization part is really cool! 
But for the project tile, I am not so sure, I think this was in an intermediate state still and I am unsure what we should actually show on the tile and if we want to sort the project list there. I asked Luca in the story, lets wait for what he says!

apps/projects/serializers.py
goapunk: merged and renamed it

apps/projects/api.py
fuzzylogic2000: Should we for the sake of speeding it up just sort by something we can more easily get for now? Like `-created`

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationProjects.jsx
```diff
@@ -21,6 +21,7 @@ export default class ModerationProjects extends Component {
     fetch(this.props.projectApiUrl)
       .then(res => res.json())
       .then(json => {
+        json.sort((a, b) => b.num_reports - a.num_reports)
```
fuzzylogic2000, line 24: Ah, and you just sort it here anyway. :)

fuzzylogic2000 (approved):
Very nice! 
I @Rineee and @philli-m are fine with it, I'd say: let's just try it on dev!

apps/userdashboard/views.py
Rineee: I will add it back in then!

fuzzylogic2000 (approved):
Nice!

PR #2348, opened 2023-04-12
notifications: add moderation specific emails

Rineee:
I actually could not test this, because I could neither make the send_test_emails work, nor could I trigger an email.. could someone check if this works? Maybe its just my setup? 
Both emails are sent to comment creator, one when comment is being blocked, the other when statement has been added to the comment..

Supposed to fix https://sentry.liqd.net/organizations/liqd/issues/3171/


goapunk:
@Rineee running `python manage.py send_test_emails admin@liqd.net` worked for me

goapunk (approved):
nice fix

Rineee:
> @Rineee running `python manage.py send_test_emails admin@liqd.net` worked for me

For me it only prints out the emails, but does not send them? So weird!

PR #2349, opened 2023-04-12
[7208] JS test (ModerationNotification)

khamui:


apps/userdashboard/assets/js/a4_candy_userdashboard/__tests__/ModerationNotification.jest.jsx
```diff
@@ -0,0 +1,90 @@
...
+  const mockedFn = jest.fn()
+  const tree = render(
+    <ModerationNotification
+      notification={mockedNotification}
+      getUrlParams={() => {}}
+      loadData={mockedFn}
+    />
+  )
+  const blockButton = tree.container.querySelector(
+    '#moderation-notification-actions-bar-button-block'
```
philli-m, line 65: just a heads up, as this is not a unique ID so I will probably re-factor in an a11y PR to ensure there is a unique identifier stuck on end, maybe pk of comment?

apps/userdashboard/assets/js/a4_candy_userdashboard/__tests__/ModerationNotification.jest.jsx
```diff
@@ -0,0 +1,90 @@
...
+test('showing highlighted alert', async () => {
+  const mockedFn = jest.fn()
+  const tree = render(
+    <ModerationNotification
+      notification={mockedNotification}
+      loadData={mockedFn}
+    />
+  )
+  const highlightButton = tree.container.querySelector(
+    '#moderation-notification-actions-bar-button-highlight'
```
philli-m, line 83: same as above

philli-m (commented):
very nice! just couple comments also maybe we should add the jest tests to `make test`and update the description?

apps/userdashboard/assets/js/a4_candy_userdashboard/__tests__/ModerationNotification.jest.jsx
khamui: yes sounds good. maybe changing, when its refactored?

khamui:
> very nice! just couple comments also maybe we should add the jest tests to `make test`and update the description?
Not sure if it is better to have py tests separated from js tests? I think it is easier to have the commands separated, so we do not need to run backend tests while working on frontend tests and vice versa. Does that make sense?

apps/userdashboard/assets/js/a4_candy_userdashboard/__tests__/ModerationNotification.jest.jsx
philli-m: added in my PR

philli-m:
> > very nice! just couple comments also maybe we should add the jest tests to `make test`and update the description?
> 
> I think it is easier to have the commands separated, so we do not need to run backend tests while working on frontend tests and vice versa. Does that make sense @philli-m ?

I would keep `make jest` and add one for py like `make test-py` and add the jest to `make test` as its useful to be able to just run them all before pushing to check, or do you think that's over complicated?

khamui:
> > > very nice! just couple comments also maybe we should add the jest tests to `make test`and update the description?
> > 
> > 
> > I think it is easier to have the commands separated, so we do not need to run backend tests while working on frontend tests and vice versa. Does that make sense @philli-m ?
> 
> I would keep `make jest` and add one for py like `make test-py` and add the jest to `make test` as its useful to be able to just run them all before pushing to check, or do you think that's over complicated?

yes maybe that is a little too complicated. In case running both tests its also possible with `make test && make jest`, so will leave it for now.

PR #2357, opened 2023-04-18
Jd 2023 04 fontawesome 6

goapunk:


goapunk:
> see a4 pr oh and check upload buttons, not sure if they work

renamed fa, far and fas as well and fixed the upload buttons

Issue #2262, opened 2023-01-31
organisation page: project tile looks tiny and short description incomplete


lthueer:
**URL:** https://adhocracy.plus/werder-havel/
**user:** 
**expected behaviour:** short description is fully shown
**behaviour:** short description incomplete
**important screensize:**
**device & browser:** 
**Comment/Question:** 
- if title is longer than one line the short description is not shown at all
- On Dev, the project tile looks very small at the moment and has a different size - something seems to be wrong.

prod
<img width="912" alt="Bildschirm­foto 2023-01-31 um 14 32 36" src="https://user-images.githubusercontent.com/36957474/215774533-9a9f5bfb-647d-46fe-97b3-746d3542a513.png">

dev
![Bildschirmfoto 2023-04-20 um 09 56 25](https://user-images.githubusercontent.com/36957474/233299061-f22b76ec-fe40-48a7-889f-410d0f810e04.png)

PR #2372, opened 2023-04-24
[2261 and 2262] project tile fixes

khamui:


philli-m (approved):
nice and good one maintaining comp lib! :) only nitpick is it doesn't seem to add the shadow to the tile images mentioned in issue #2262 but if that's a much longer fix then feel free to merge

Issue #2364, opened 2023-04-20
categories, labels, Status look off


lthueer:
**URL:** https://aplus-dev.liqd.net/teststadt/ideas/2022-00733/
**user:** 
**expected behaviour:** categories, labels, status look as on prod
**behaviour:** categories, labels, status look small. For the status accepted, rejected etc. the colour is missing
**important screensize:**
**device & browser:** 
**Comment/Question:** 
- has this been changed on purpose? I think before they were quite big but this def looks too small to me

![Bildschirmfoto 2023-04-20 um 10 15 15](https://user-images.githubusercontent.com/36957474/233304085-675bd23b-64da-4ad8-81e5-8e89e09da8b2.png)

![Bildschirmfoto 2023-04-20 um 10 15 19](https://user-images.githubusercontent.com/36957474/233304110-a1b1715b-8bff-49ee-bc8b-3dfe92cafac3.png)

debate module
<img width="851" alt="Bildschirmfoto 2023-04-24 um 15 57 32" src="https://user-images.githubusercontent.com/36957474/234018722-a6d4fa9f-8256-41c1-a380-30858ca20709.png">

lthueer:
@philli-m @khamui just a note: the issue is still there for the debate module (last screenshot). If it's jsut not on dev yet, please ignore my comment

khamui:
@lthueer seems to be still an issue. having a look now

lthueer:
@khamui @philli-m this issue seems to be fixed for the debate module but not for the labels and the traffic light status in brainstorming modules: https://aplus-dev.liqd.net/teststadt/projects/wettbewerb-der-ideen-in-musterstadt/ Can someone have a look at this? :)

khamui:
this is really weird, as this worked. I will have a look again.

lthueer:
Thank you @khamui

lthueer:
@SandisStudi I think in the interactive event its fine like this. Before this issue came up it was not as big as the labels in Brainstorming. I would suggest we keep it like that! Thanks for pointing that out.

PR #2381, opened 2023-04-26
badge.scss: making sure bs and a4 classes are not overwritten by .badge

khamui:
this fixes the badge issue in debate module, mentioned by @lthueer here:
https://github.com/liqd/adhocracy-plus/issues/2364#issuecomment-1521960083


philli-m (approved):
nice fix, however I think this is actually an issue that we misuse the `badge` class in the comment categories as they are in fact labels, can you add a comment to this line to FIXME in a4? and hopefully we will remember next time we touch this bit of a4. After feel free to merge!

PR #2383, opened 2023-04-26
userdashboard/views: exclude paragraphs from comment actions

Rineee:
This should fix https://sentry.liqd.net/organizations/liqd/issues/3179/


philli-m (approved):
looking good! will wait for tests to pass

PR #2385, opened 2023-04-26
update a4 hash

### LINKED ISSUE #2284: user agreement state not updated
**URL:** 
**user:** 
**expected behaviour:** when I add a comment and agree to terms of use, I dont have to agree again when I add a reply to another comment on the same page
**behaviour:** the state is not updated (or passed) and I have to tick the box again when adding a reply to another comment
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?

khamui:
to receive the fix #2284


fuzzylogic2000 (approved):
Cool!

PR #2390, opened 2023-04-27
apps/userdashboard/ModerationNotificationList: fix packet size when f…

### LINKED ISSUE #2350: #7211 loaded comments pagination (load more) after filtering
**URL:** https://aplus-dev.liqd.net/userdashboard/moderation/detail/jonathan-app-test/
**user:** moderator, initiator
**expected behaviour:** as I user I want to have only 15 comments loaded when I change the filter.
**behaviour:** as a user when I already loaded more comments and then change filter it shows still more then 15 comments. 
**important screensize:** any
**device & browser:** tested on mac (firefox)
**Comment/Question:** I talked to Luca and he said, that after the filter is changed the amount of loaded comments should be again limited to 15.
 
Screenshot?

khamui:
…iltering

fixes #2350


Rineee (approved):
Cool, works really nicely!

PR #2391, opened 2023-04-27
[#issues] front end release issues

philli-m:


adhocracy-plus/assets/scss/components/_button.scss
```diff
@@ -283,7 +283,7 @@
     border: 0;
 }
 
-.checkbox-btn__badge--light {
+.checkbox-btn__label--light {
```
philli-m, line 286: @khamui yo this was my bad with the review, these elements are actually labels so their naming is correct, sorry for the mess up!

goapunk (approved):
looks good to me!

PR #2398, opened 2023-05-02
apps/users: adding "account already exists" email

### LINKED ISSUE #2271: TemplateDoesNotExist: account/email/account_already_exists.en.email
Sentry Issue: [APLUS-PROD-33](https://sentry.liqd.net/organizations/liqd/issues/2941/?referrer=github_integration)

```
TemplateDoesNotExist: account/email/account_already_exists.en.email
(16 additional frame(s) were not displayed)
...
  File "apps/users/adapters.py", line 55, in send_mail
    return UserAccountEmail.send(email, template_name=template_prefix, **context)
  File "adhocracy4/emails/mixins.py", line 38, in send
    return cls().dispatch(object, *args, **kwargs)
  File "adhocracy4/emails/base.py", line 127, in dispatch
    (subject, text, html) = self.render(template, context)
  File "apps/users/emails.py", line 60, in render
    template = get_template(template_name + ".en.email")
  File "django/template/loader.py", line 19, in get_template
    raise TemplateDoesNotExist(template_name, chain=chain)
```

khamui:
fixes #2271

### DIFF
--- a/adhocracy-plus/templates/account/email/account_already_exists.en.email
+++ b/adhocracy-plus/templates/account/email/account_already_exists.en.email
@@ -1,0 +1,21 @@
+{% extends 'email_base.'|add:part_type %}
+{% load i18n %}
+
+{% block subject %}{% blocktranslate %}Account Already Exists{% endblocktranslate %}{% endblock %}
+
+{% block content %}
+<p>
+    {% blocktranslate %}You are receiving this e-mail because you or someone else tried to signup for an
+    account using the e-mail address:{% endblocktranslate %}
+</p>
+<p>
+    {{ email }}
+</p>
+<p>
+    {% blocktranslate %}However, an account using that e-mail address already exists.  In case you have
+    forgotten about this, please use the password forgotten procedure to recover
+    your account:{% endblocktranslate %}
+</p>
+{% endblock %}
+{% block cta_url %}{{ password_reset_url }}{% endblock %}
+{% block cta_label %}{% blocktranslate %}Reset password{% endblocktranslate %}{% endblock %}
--- a/apps/users/forms.py
+++ b/apps/users/forms.py
@@ -63,10 +63,11 @@ class DefaultSignupForm(SignupForm):
 
     def save(self, request):
         user = super().save(request)
-        user.get_newsletters = self.cleaned_data["get_newsletters"]
-        user.language = get_language()
-        user.save()
-        return user
+        if user:
+            user.get_newsletters = self.cleaned_data["get_newsletters"]
+            user.language = get_language()
+            user.save()
+            return user
 
 
 class IgbceSignupForm(DefaultSignupForm):
@@ -97,7 +98,6 @@ class IgbceSignupForm(DefaultSignupForm):
     )
 
     def validateMemberNumberAndDate(self, member_number, birth_date):
-
         if not hasattr(settings, "IGBCE_NAV_URL") or not hasattr(
             settings, "IGBCE_NAV_SECURITYID"
         ):
@@ -167,7 +167,6 @@ class IgbceSignupForm(DefaultSignupForm):
         self.validateMemberNumberAndDate(member_number, birth_date)
 
     def save(self, request):
-
         user = super().save(request)
         if hasattr(settings, "SITE_ORGANISATION_SLUG"):
             organisation = Organisation.objects.get(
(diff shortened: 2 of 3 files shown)


apps/users/forms.py
```diff
@@ -63,10 +63,11 @@ def __init__(self, *args, **kwargs):
...
         user = super().save(request)
-        user.get_newsletters = self.cleaned_data["get_newsletters"]
-        user.language = get_language()
-        user.save()
-        return user
+        if user:
+            user.get_newsletters = self.cleaned_data["get_newsletters"]
+            user.language = get_language()
+            user.save()
+            return user
```
Rineee, line 70: Hm, would the return outside the if statement also work? It somehow looks strange to me, that in one case the function returns sth and the other not..

adhocracy-plus/templates/account/email/account_already_exists.de.email
```diff
@@ -0,0 +1,20 @@
...
+     {{ email }}
+ </p>
+ <p>
+     Es existiert jedoch bereits ein Konto mit dieser E-Mail-Adresse.
+     Falls Sie dies vergessen haben, verwenden Sie bitte das Verfahren "Passwort vergessen",
+     um Ihr Konto wiederherzustellen:
+ </p>
+ {% block cta_url %}{{ password_reset_url }}{% endblock %}
+ {% block cta_label %}Passwort zurücksetzen{% endblock %}
+ {% endblock %}
```
Rineee, line 20: We normally translate the emails, so we would only need the english version, but with `{% blocktranslate %}` tags around, like in the other emails.

Rineee:
Cool! Although I cannot test, because my emails are not working locally :crying_cat_face: Maybe someone else can check that?

adhocracy-plus/templates/account/email/account_already_exists.en.email
```diff
@@ -0,0 +1,21 @@
...
+</p>
+<p>
+    {{ user.email }}
+</p>
+<p>
+    {% blocktranslate %}However, an account using that e-mail address already exists.  In case you have
+    forgotten about this, please use the password forgotten procedure to recover
+    your account:{% endblocktranslate %}
+</p>
+{% block cta_url %}{{ password_reset_url }}{% endblock %}
```
goapunk, line 19: The cta block is not part of the content block, it renders incorrectly in the email. If you move the endblock from line 21 to here it works fine :)

goapunk (changes requested):
looks good, just a minor fix in the template

adhocracy-plus/templates/account/email/account_already_exists.en.email
```diff
@@ -0,0 +1,21 @@
...
+
+{% block subject %}{% blocktranslate %}Account Already Exists{% endblocktranslate %}{% endblock %}
+
+{% block content %}
+<p>
+    {% blocktranslate %}You are receiving this e-mail because you or someone else tried to signup for an
+    account using e-mail address:{% endblocktranslate %}
+</p>
+<p>
+    {{ user.email }}
```
fuzzylogic2000, line 12: This does not put my email into the mail, it should be `{{ email }}`

fuzzylogic2000 (commented):
And a third one. :) The E-Mail sending is very slow in chrome, but works fine in FF (?). That's hopefully fine once it's on the server. 
But the email address doesn't get into email.

adhocracy-plus/templates/account/email/account_already_exists.en.email
khamui: @fuzzylogic2000 for me it only works with `user.email`. `email` only, shows the memory address

goapunk:
> And a third one. :) The E-Mail sending is very slow in chrome, but works fine in FF (?). That's hopefully fine once it's on the server. But the email address doesn't get into email.

It seems to load forever in firefox as well. Maybe we should investigate, because it seems unreasonably slow

goapunk (approved):
nice :)

khamui:
@fuzzylogic2000 @goapunk for me (in chrome) it was "fast", not specially slow. I am not sure if this is related to browsers. Shall we make a separate issue? Or maybe either @philli-m or @Rineee could also check and if it is slow for you too, we immediately investigate?

goapunk:
@Rineee @fuzzylogic2000 do you want to review again or ready to merge?

Rineee:
Im fine merging

PR #2401, opened 2023-05-02
apps//userdashboard//projects: fix tiles clickable and shadow (multim…

khamui:
…odule projects and moderation dashboard).

Note: this is related to an issue fix, which was already closed.

### DIFF
--- a/apps/projects/templates/a4_candy_projects/includes/project_list_tile.html
+++ b/apps/projects/templates/a4_candy_projects/includes/project_list_tile.html
@@ -1,11 +1,14 @@
 {% load i18n project_tags a4_candy_project_tags contrib_tags thumbnail rules static %}
 
 <li class="tile tile--{{ orientation }}">
-
+    <a class="tile__link" aria-labelledby="tile-title-{{ object.pk }}" href="{{ url }}">
+        <h3 class="visually-hidden" id="tile-title-{{ object.pk }}">
+            {{ object.name }}
+        </h3>
+    </a>
     {% if type == 'project' %}
     {% project_tile_image object as project_image %}
     {% project_tile_image_copyright object as project_image_copyright %}
-    <a class="tile__link" aria-labelledby="tile-title-{{ object.pk }}" href="{{ url }}"></a>
     <div class="tile__head">
         <div class="tile__image"
           {% if project_image %}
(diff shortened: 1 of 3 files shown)


apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationProjects.jsx
```diff
@@ -105,13 +111,7 @@ export default class ModerationProjects extends Component {
...
                       <span className="text-muted">{byText}{item.organisation}</span>
                       <h3 className="tile__title mb-2">
-                        <a
-                          href={item.moderation_detail_url}
-                          className="tile__link"
-                        >
-                          {item.num_unread_comments > 0 && <span className="text-info">• <span className="visually-hidden">{hasUnReadComments}</span></span>}
-                          {item.title}
-                        </a>
+                        {item.title}
```
goapunk, line 114: one thing I noticed is that it is now read differently by the Screenreader (at least the chrome one). It doesn't read the title anymore when  tabbing through (so you need an extra step to read it). Not sure if that's an issue though. @philli-m any thoughts?

goapunk (approved):
apart from the comment about the SR looks good to me

apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationProjects.jsx
philli-m: I can check on JAWS and NVDA as they are the most used

goapunk: @khamui I used the one the chrome screen reader (which probably isn't very good). I can check with the other ones later.

goapunk: NVDA only reads blank, I guess that's not ideal

khamui: @goapunk @philli-m having the visually hidden h3 as a child worked finally..

apps/projects/templates/a4_candy_projects/includes/project_list_tile.html
```diff
@@ -1,11 +1,14 @@
 {% load i18n project_tags a4_candy_project_tags contrib_tags thumbnail rules static %}
 
 <li class="tile tile--{{ orientation }}">
-
+    <a class="tile__link" aria-labelledby="tile-title-{{ object.pk }}" href="{{ url }}">
+        <h3 class="visually-hidden" id="tile-title-{{ object.pk }}">
+            {{ object.name }}
+        </h3>
+    </a>
```
philli-m, line 8: @khamui that took me too long to spot :p the link was just inside the if for tile image, so tiles with no image didn't have it

philli-m (approved):
nice!

PR #2416, opened 2023-05-09
userdashboard: remove unused fetch of comment count

goapunk:


apps/userdashboard/views.py
```diff
@@ -46,9 +45,6 @@ def projects(self):
         projects = Project.objects.filter(
             follow__creator=self.request.user, follow__enabled=True
         )
-        for project in projects:
-            project.latest_comments_num = helpers.get_num_latest_comments(project)
```
goapunk, line 50: I couldn't  find any place we use it, but it makes the site load very slow

fuzzylogic2000: You are right, we don't use it!

fuzzylogic2000 (approved):

PR #2449, opened 2023-06-06
Jd 2023 06 userdashboard filter 2

goapunk:
depends on https://github.com/liqd/adhocracy4/pull/1421


apps/userdashboard/views.py
```diff
@@ -86,16 +86,15 @@ def actions(self):
                 obj_content_type=ContentType.objects.get_for_model(
                     ModeratorCommentFeedback
                 ),
+                obj_creator=user,
```
Rineee, line 89: Should this be the feedback_comment_creator?

goapunk: ups :)

Rineee (approved):
yeah!

PR #2450, opened 2023-06-06
projects/helpers: simplify comment filter

goapunk:
- [x] depends on https://github.com/liqd/adhocracy4/pull/1422

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "file-saver": "2.0.5",
     "glob": "10.2.7",
     "immutability-helper": "3.1.1",
-    "jquery": "3.6.4",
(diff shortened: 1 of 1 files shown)


apps/projects/helpers.py
```diff
@@ -11,24 +11,7 @@
 
 def get_all_comments_project(project):
     return Comment.objects.filter(
-        Q(budget_proposal__module__project=project)
```
goapunk, line 14: the individual filtering here is really slow, do we really need it? what other comments are there which we don't want to show here?

philli-m (approved):
looking good, can't really test as my database is very small but we can check on dev

PR #2486, opened 2023-07-04
[7428] templates/module_detail: pass initial_slide in link back to project

Rineee:
This needs https://github.com/liqd/adhocracy4/pull/1443

@m4ra @hklarnerliqd if you want to test it, you can checkout locally the a4 pull request and link it to your local adhocracy plus, like described here: https://github.com/liqd/adhocracy4/blob/main/docs/development.md#development-setup


hklarner:
looks good to me

PR #2489, opened 2023-07-11
WIP [7431] apps/projects: add ratings and comments insight

m4ra:
Starting with counting comments and ratings of all related project's modules.
I included a commented block as well to decide on which path to continue.

### DIFF
--- a/apps/projects/templates/a4_candy_projects/project_detail.html
+++ b/apps/projects/templates/a4_candy_projects/project_detail.html
@@ -279,24 +279,13 @@
                 aria-expanded="false">
                 <div class="container">
                     <div class="row">
-                        <div class="col-md-10 col-lg-8 offset-md-1 offset-lg-2">
-                            {% if comments %}
-                                Comments: {{ comments }}
-                            {% else %}
-                                {% translate 'No comments yet.' %}
-                            {% endif %}
 
-                        </div>
-                    <div class="row">
-                        <div class="col-md-10 col-lg-8 offset-md-1 offset-lg-2">
-                            {% if ratings %}
-                                Ratings: {{ ratings }}
-                            {% else %}
-                                {% translate 'No ratings yet.' %}
-                            {% endif %}
+                        {% for label, value in counts %}
+                            <div class="col-md-10 col-lg-8 offset-md-1 offset-lg-2">
+                                {{ label }}: {{ value }}
+                            </div>
+                        {% endfor %}
 
-                        </div>
-                    </div>
                     <div class="row">
                         <div class="col-md-10 col-lg-8 offset-md-1 offset-lg-2">
                             {% if project.result %}
--- a/apps/projects/templates/a4_candy_projects/project_detail.html
+++ b/apps/projects/templates/a4_candy_projects/project_detail.html
@@ -280,6 +280,25 @@
                 <div class="container">
                     <div class="row">
                         <div class="col-md-10 col-lg-8 offset-md-1 offset-lg-2">
+                            {% if comments %}
+                                Comments: {{ comments }}
+                            {% else %}
+                                {% translate 'No comments yet.' %}
+                            {% endif %}
+
+                        </div>
+                    <div class="row">
+                        <div class="col-md-10 col-lg-8 offset-md-1 offset-lg-2">
+                            {% if ratings %}
+                                Ratings: {{ ratings }}
+                            {% else %}
+                                {% translate 'No ratings yet.' %}
+                            {% endif %}
+
+                        </div>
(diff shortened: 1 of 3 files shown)


m4ra:
@hklarner is ready to fetch.

apps/projects/views.py
```diff
@@ -291,9 +292,55 @@ def is_project_view(self):
...
+
+        modules = self.project.modules.all()
+        ideas = Idea.objects.filter(module__in=modules)
+        map_ideas = MapIdea.objects.filter(module__in=modules)
+        proposals = Proposal.objects.filter(module__in=modules)
+        comments = get_all_comments_project(project=self.project)
+
+        ratings_ideas = Rating.objects.filter(idea__in=ideas)
+        ratings_map_ideas = Rating.objects.filter(mapidea__in=map_ideas)
+        ratings_comments = Rating.objects.filter(comment__in=comments)
```
m4ra, line 307: @hklarner here you are hitting the DB again, that is why I had the prefetch in the comments with the ratings, and also now you are including the ratings that have value 0. Why would you remove the `Ratings.objects.exclude(0)`?
ref https://docs.djangoproject.com/en/3.2/ref/models/querysets/#prefetch-related

apps/projects/templates/a4_candy_projects/project_detail.html
```diff
@@ -278,6 +278,14 @@ <h1 class="project-header__title">{{ module.name }}</h1>
                 aria-labelledby="tab-project-{{ project.pk }}-result"
                 aria-expanded="false">
                 <div class="container">
+                    <div class="row">
+
+                        {% for label, value in counts %}
+                            <div class="col-md-10 col-lg-8 offset-md-1 offset-lg-2">
+                                {{ label }}: {{ value }}
+                            </div>
+                        {% endfor %}
```
m4ra, line 287: Nice!

apps/projects/views.py
```diff
@@ -294,49 +298,87 @@ def raise_exception(self):
...
+        live_questions = LiveQuestion.objects.filter(module__in=modules)
+        likes = Like.objects.filter(livequestion__in=live_questions)
+
+        values = [Rating.POSITIVE, Rating.NEGATIVE]
+        ratings_ideas = Rating.objects.filter(idea__in=ideas, value__in=values)
+        ratings_map_ideas = Rating.objects.filter(
+            mapidea__in=map_ideas, value__in=values
+        )
+        ratings_comments = Rating.objects.filter(comment__in=comments, value__in=values)
+
```
Rineee, line 330: The ratings on topics are missing (topic prioritization module), probably also a thing to add below with the flags?

apps/projects/views.py
```diff
@@ -294,49 +298,87 @@ def raise_exception(self):
...
+            count_ideas += ideas.count()
+            count_ratings += ratings_ideas.count()
+
+        if enable_map_ideas:
+            creator_objects.extend([map_ideas, ratings_map_ideas])
+            count_ideas += map_ideas.count()
+            count_ratings += ratings_map_ideas.count()
+
+        if enable_polls:
+            count_poll_answers += votes.count() + answers.count()
```
Rineee, line 353: I think here we also need to add votes and answers to creator_objects.

apps/projects/views.py
```diff
@@ -294,49 +298,87 @@ def raise_exception(self):
 
     def get_context_data(self, **kwargs):
         """Append insights to the template context."""
+
         context = super().get_context_data(**kwargs)
+        idea_blueprints = {"BS", "IC"}
+        map_idea_blueprints = {"MBS", "MIC", "PB"}
```
Rineee, line 304: The participatory budgeting needs to be treated separately as well, there Proposal model is used. They also contribute to the idea count though.

Rineee:
@hklarner Cool, looking good. I added some comments in the code.. and should we maybe move the counting to a separate function, what do you think, also @m4ra ?

m4ra:
Yes the countings can move to project/helpers too along with the other queries. And add tests @hklarner @Rineee

apps/projects/views.py
hklarner: btw the `.filter()` is lazy, it doesn't go against the DB, see
- https://docs.djangoproject.com/en/4.2/topics/db/queries/#querysets-are-lazy

apps/projects/views.py
hklarner: True, will add them.

apps/projects/views.py
hklarner: yes! will do

PR #2491, opened 2023-07-11
assets/fonts: add logo as icon //icons: set icon classname and style …

philli-m:
…//button: add colour and hover styling


goapunk (approved):
looks good to me

Issue #2392, opened 2023-04-27
#7208 [a+]: Integrate new User and Moderation Dashboard from Kosmo


mcastro-lqd:
**URL:** https://aplus-dev.liqd.net/userdashboard/moderation/
**user:** *moderator/ initiator/ participants*
**important screensize:** *Both*
**device & browser:** *Safari Version 16.1 (18614.2.9.1.12) & Google Chrome Version 112.0.5615.137 (Official Build) (x86_64)*
**Comment/Question:** *There was so many small things that I decided to put everything on a pdf instead of creating an issue for each, If any of these are too big, I can also write a separated issue for that. Let me know if anything is unclear. Thanks in advance. * 

[screencapture-aplus-dev-liqd-net-userdashboard-moderation-2023-04-26-17_30_45.pdf](https://github.com/liqd/adhocracy-plus/files/11346436/screencapture-aplus-dev-liqd-net-userdashboard-moderation-2023-04-26-17_30_45.pdf)

mcastro-lqd:
@philli-m Id just leave it then

PR #2494, opened 2023-07-13
[issues] release issue fixes

philli-m:


adhocracy-plus/assets/scss/icons/style.css
```diff
@@ -81,6 +82,10 @@
     content: "\e90a";
 }
 
+.icon__img {
```
goapunk, line 85: instead of using css I think we could just change the thumbnail tag from `0x40` to `0x32`

apps/userdashboard/templates/a4_candy_userdashboard/base_userdashboard.html
```diff
@@ -76,9 +76,10 @@ <h2 class="userdashboard-profile__name">
                       <h2 class="fs-6">{% trans 'Initiator Organisations' %}</h2>
                       {% for organisation in request.user.organisations %}
                       <div>
-                          <a class="btn btn--link text-decoration-none text-muted d-flex align-items-center" href="{% url 'a4dashboard:project-list' organisation_slug=organisation.slug %}">
+                          <a class="btn btn--link text-decoration-none u-text--gray d-flex align-items-center" href="{% url 'a4dashboard:project-list' organisation_slug=organisation.slug %}">
                             {% if organisation.logo %}
                               <img
+                                class="icon__img"
                                 src="{% thumbnail organisation.logo '0x40' %}"
```
goapunk, line 83: `0x40` -> `0x32`
same for the other ones below

goapunk (changes requested):
looks really nice, just a small comment on the thumbnails and the vertical space and breakpoint in the dashboard (see https://github.com/liqd/adhocracy-plus/issues/2392)

apps/userdashboard/templates/a4_candy_userdashboard/base_userdashboard.html
goapunk: if there is no icon the font awesome icon is still 40px, should that be changed as well?

philli-m: no i think that should be ok but will see on dev when there multiple orgs

philli-m: ok have to add new icon styling but all fixed now

PR #2498, opened 2023-07-20
[7428] tests: add test for multi-module timeline module detail back button

goapunk:
test that the back button in the module detail view gets passed the correct initial slide parameter


philli-m (approved):
LGTM

PR #2513, opened 2023-08-07
docs/project_insights: add structure and auto-format

goapunk:


docs/project_insights.md
```diff
@@ -1,13 +1,41 @@
+## Structure
+
+### frontend
```
goapunk, line 3: not sure if the separation is helpful

hklarner:
looks good to me

PR #2516, opened 2023-08-08
[7431] fixes project insights for draft modules

hklarner:
- without this fix: results tab shows optional insights (polls, live questions, ideas) for modules that are in draft state
- with this fix: only non-draft modules can trigger optional insights


goapunk (approved):
LGTM

PR #2518, opened 2023-08-09
update translations

goapunk:


locale-source/locale/tr/LC_MESSAGES/djangojs.po
```diff
@@ -1126,8 +1134,8 @@ msgstr ""
 
 #: apps/userdashboard/assets/js/a4_candy_userdashboard/ModerationNotification.jsx:221
 #, javascript-format
-msgid "'s {}comment{} has been reported 1 time since it's creation"
-msgid_plural "'s {}comment{} has been reported %s times since it's creation"
+msgid "This {}comment{} has been reported 1 time since it's creation"
```
m4ra, line 1137: I think correct grammar is `its creation` or `it's created`. Can you check with @philli-m ?

m4ra (changes requested):
there is a grammar err.

m4ra (approved):

m4ra:
@goapunk aren't we squashing commits into one? Do we need a fix-typo commit separately in the git history? Also @hklarner has a similar commit in his PR.

goapunk:
> @goapunk aren't we squashing commits into one? Do we need a fix-typo commit separately in the git history? Also @hklarner has a similar commit in his PR.

@m4ra generally (I think) we only squash if commits are part of one (connected) change/feature. But if they are logically separate changes they should have their own commit (and then can also be easily reverted individually if necessary). In this particular case the changes in the second commit are auto-generated by django / transifex, so I wouldn't mix them with my manual changes.

PR #2519, opened 2023-08-09
tests: removes __init__.py files from root subdirectories

hklarner:
- `__init__.py` files are not required by pytest
- `__init__.py` in rules folders lead to crashes of PyCharm debugger


goapunk (approved):
nice!

PR #2522, opened 2023-08-18
[7429] optimize navigation for idea/topic lists

m4ra:


apps/projects/helpers.py
```diff
@@ -41,3 +43,12 @@ def get_num_reported_unread_comments(project):
         .filter(num_reports__gt=0)
         .count()
     )
+
+
+def get_pagination(obj: object, project: object, per_page=15) -> int:
+    objects = obj.__class__.objects.filter(module=project)
```
hklarner, line 49: rename parameter `project` to `module` in function `get_pagination`. Either it is a project or a module, but `filter(module=project)` looks wrong.

hklarner:
Nice! I did some manual testing and it works: the back button takes you back to the page you came from.

apps/projects/helpers.py
```diff
@@ -41,3 +43,12 @@ def get_num_reported_unread_comments(project):
...
+    p = Paginator(objects, per_page)
```
hklarner, line 50: tbh I don't think we need the `Paginator` here since you are computing the page number yourself. All you need is the index of `obj` in the `objects` query set. The `p.object_list` is literally just the objects query set you put in. You could equally simply iterate yourself:
```python
index = None
for i, other_object in enumerate(objects):
    if other_object == obj:
        index = i + 1

if index is None:
    # error case, obj not in objects
    page_number = 1
else:
    page_number = ceil(index / per_page)
```

apps/ideas/views.py
```diff
@@ -84,6 +85,11 @@ class IdeaDetailView(AbstractIdeaDetailView):
     )
     permission_required = "a4_candy_ideas.view_idea"
 
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        context["page_number"] = get_pagination(self.get_object(), self.module)
```
hklarner, line 90: I think instead of using a hard-coded default for how many items there are on a page (`per_page=15`) I would use the value already defined for the `AbstractIdeaListView`, so that if you change it, the new page number function will still compute the correct value. So maybe something like this:
```python
def get_context_data(self, *args, **kwargs):
    context = super().get_context_data(*args, **kwargs)

    context["page_number"] = get_page_number(
        object=self.get_object(),
        objects=self.model.objects.filter(module=self.module),
        paginate_by=AbstractIdeaListView.paginate_by,
    )

    return context
```

Also:
- I would compute the `objects` outside of `get_page_number` to avoid the `obj.__class__.objects`
- I would rename `per_page` to `paginate_by` to be in agreement with the respective parameter of `AbstractIdeaListView`

m4ra: It would be good, but we want to use this function for other modules too, e.g topics.

hklarner: The `AbstractIdeaListView` is inherited by several views, including the `TopicListView`:
```
ProposalListView
SubjectListView
IdeaListView
MapIdeaListView
TopicListView
```

But now that you mention it, shouldn't the new context then be part of the `AbstractIdeaDetailView` instead of `IdeaDetailView`? (so that it also works for topics, map ideas etc. by inheritence?)

apps/contrib/templates/a4_candy_contrib/item_detail.html
```diff
@@ -17,10 +17,19 @@
                 <nav class="breadcrumbs" aria-label="{% translate 'breadcrumbs' %}">
                     <ul>
                         <li>
-                            <a href="{{ module.get_detail_url }}">
+                          {% with navigate_url=request.META.HTTP_REFERER %}
+                            {% if "?" in navigate_url %}
```
goapunk, line 21: I think the condition here is not working as intended. From how I understand the story there are two cases:

1. you open a project and have list of ideas. You click on an idea to open the detail view. In this case we want to show the `go back` button which will take you back to the list with correct page/pagination and filters set
2. someone shares a link to an idea. In this case we want to show the `go to overview` button and just go back without pagination/filtering.

By checking for ? we actually show the `go back to overview` if you click on any idea on page > 1,  a non-default sorting or a filter is set.  Opening an idea from a link will show `go back`  instead of `go back to overview`.

goapunk (changes requested):

apps/contrib/templates/a4_candy_contrib/item_detail.html
m4ra: I actually understood it as I have implemented it. When we go to a detail view from the pagination we go back to the overview of the ideas, when we share the link we go back to the project page. I will look at the story again.

m4ra: So the `only when coming from link: go to overview` means when the link is shared, that wasn't clear. I will change it accordingly. The qn now is, should both cases navigate to the part of the page with the items list? so the `go to overview` should also take the user to the default ordering/filtered list? Or `go to overview` takes back at the header/title of the project, meaning the `{{navigate_url}}`without the `index` anchor?

goapunk: hmm, not sure, I think going to the index/list instead to the top of the page when coming from a link would be fine but we can also ask @lthueer what he prefers

m4ra: I will leave it now going back to the index anchor, we can always change it if that is not the case.

m4ra: @goapunk fixed. I now check for the module's url in the HTTP-referer, which is enough and safer to distinguish between the 2 cases.

goapunk (changes requested):
code looks good but I think the debate module is missing :)

m4ra:
@goapunk indeed was missing, added it now, thanx!

goapunk (approved):
works great!

PR #2527, opened 2023-09-18
refactors apps/debate and renames image fixtures

hklarner:
The intention of this refactor is to
- rename the image fixtures to adhere to Python naming conventions (camel case)
- simplify the property `last_three_creators` by filtering the query directly

### DIFF
--- a/apps/debate/models.py
+++ b/apps/debate/models.py
@@ -73,14 +73,21 @@ class Subject(module_models.Item):
 
     @cached_property
     def last_three_creators(self):
-        comments = self.comments.all().select_related("creator").order_by("-created")
+        comments = (
+            self.comments.filter(
+                is_censored=False,
+                is_removed=False,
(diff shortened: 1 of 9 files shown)


apps/debate/models.py
```diff
@@ -73,14 +73,15 @@ def comment_creator_count_minus_three(self):
 
     @cached_property
     def last_three_creators(self):
-        comments = self.comments.all().select_related("creator").order_by("-created")
+        comments = self.comments.filter(is_censored=False, is_removed=False)
+        comments = comments.select_related("creator").order_by("-created")
+
```
m4ra, line 78: nice. and just do the filter and select_related in one query. we can have new lines in one queryset for every `.`

hklarner: done

m4ra (approved):

PR #2542, opened 2023-11-09
[7638] Drop background_task tables

m4ra:
**Tasks**
- [x] Changelog


apps/contrib/migrations/0001_drop_background_task_table.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 3.2.20 on 2023-11-09 07:46
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = []
+
+    operations = [
+        migrations.RunSQL(
```
goapunk, line 10: the signature of `RunSQL` is `class RunSQL(sql, reverse_sql=None, state_operations=None, hints=None, elidable=False)`, so this doesn't actually drop both tables but sets the reverse_sql. We need to wrap them in square brackets. Sorry, bad reviewing on my side

goapunk (changes requested):

goapunk (approved):
LGTM

PR #2552, opened 2024-02-05
[7637] django upgrade to 4.2

m4ra:
To be merged after wagtail upgrade [#2554](https://github.com/liqd/adhocracy-plus/pull/2554/)

**Tasks**
- [x] Changelog


m4ra:
@goapunk there is an issue with the background task app. So I will rebase this PR once you merge https://github.com/liqd/adhocracy-plus/pull/2542 into main.

goapunk (changes requested):
Nice, looks good, just one thing which I think is missing: the delete() -> form_valid() change was part of this update already

goapunk (approved):
Really nice! Only thing we need to take into account is that by updating the a4 hash we now have the ckeditor5 changes, so we can't release this until we did the ckeditor5 story

m4ra:
@goapunk or we don't update the a4 hash.

goapunk:
> @goapunk or we don't update the a4 hash.

will it cause problems with a4 on django 3.2 and a+ on django 4.2 if we don't update the hash?

m4ra:
@goapunk the topics might be an issue, though we don't use it in a+. Need to check if ckeditor was added after the multiselect-field.

goapunk (approved):
LGTM, we didn't do the deprecated app config bit but we can do them in a new PR and get this merged I'd say

Issue #2538, opened 2023-11-05
Need help in adding a new module without start-time and end-time


erelsgl:
Hi all. We want to add a new module, in which users can both add ideas and rank them at the same time, without any time-limit.
It is similar to the "idea challenge" module, except that adding and ranking occurs simultaneously, and there are no start-time and end-time.

We looked at the "idea challenge" module to get an idea what to do, but got stuck. Particularly, we did not see any reference to "start-time" or "end-time" in [apps.ideas.phases](https://github.com/liqd/adhocracy-plus/blob/main/apps/ideas/phases.py).

* Where are the "start-time" and "end-time" kept? How are they stored and restored?
* What should we do in our new module, to remove the requirement for a start-time and end-time?
* What should we do to allow ranking and adding ideas simultaneously?

We are new to this project, so any explanations or references to the right locations in the code are welcome.
Thanks!

goapunk:
Hi @erelsgl ,
thanks for your interest in a+. Creating a module which allows you to add and rank ideas at the same time should be fairly easy. We already have a phase which allows that (see [here](https://github.com/liqd/adhocracy-plus/blob/main/apps/ideas/phases.py#L63)), so all you need to do is to create a blueprint which uses this phase, see  [blueprints](https://github.com/liqd/adhocracy-plus/blob/main/apps/dashboard/blueprints.py). In fact, for testing you can even create a module with that features already without any modifications if you are a bit familiar with Django and its admin interface. To do so these are the steps roughly:

1. Create a project in the dashboard
2. Go to the module pane in the django admin interface (e.g. http://127.0.0.1:8004/django-admin/a4modules/module/)
3. Click `Add Module`
4. Fill in **all** fields, for weight choose 1, for project choose the project you created in step 2. For the phases part, fill in all fields and for `type` choose `CollectFeedbackPhase`. 
5. Go back to dashboard and publish your project

Regarding the time-limit, creating a phase without time constraints might unfortunately be more complicated as a lot of the logic in the code relies on the dates being present. A trivial workaround which doesn't require any code modifications would be to just set the end-date in the very far future, which would effectively be the same as without a time-limit. If you want to have a look into changing the code I recommend looking at [adhocracy4](https://github.com/liqd/adhocracy4), our library which is the base for a+ and our other projects. This is also where the phases are defined, see  [here](https://github.com/liqd/adhocracy4/blob/main/adhocracy4/phases/models.py#L63). 
If you have any more questions feel free to ask.

[...]

PR #2555, opened 2024-02-14
fair-vote module

oriyalperin:
Hi,
regarding issue #2538,
we've added an initial implementation of our new module, as detailed [here](https://github.com/ariel-research/adhocracy-plus/blob/fairvote/FAIRVOTE.md).
We would appreciate your comments on the idea itself and any suggestions on how we can improve the implementation.
Thank you!


goapunk:
hi @oriyalperin , thanks for the pull request and the idea :) I can't promise we'll be able to give any (detailed) feedback as we are currently quite busy but I will hopefully have the time to have a look at some point

PR #2559, opened 2024-02-28
[8046] Update ckeditor5 django

m4ra:
This PR still is missing the ckeditor5 changes for assets/<filename>.jsx

TODO:

- [ ] react documents
- [x] iframe migrations for organisation
- [x] iframe migration for offlineevent
- [x] iframe migration for interactiveevent

**Tasks**
- [ ] Changelog


apps/documents/assets/ParagraphForm.jsx
```diff
@@ -82,12 +97,19 @@ class ParagraphForm extends React.Component {
...
+                {translations.paragraph}
+                <div
+                  id={'id_paragraph-help-text-' + props.id}
+                  className="form-hint"
+                >
+                  {translations.helpText}
+                </div>
                 <div
                   className="django-ckeditor-widget"
+                <div
```
goapunk, line 110: something is wrong here `<div [...] div<`

m4ra: oh totally... i updated it.

apps/documents/templatetags/react_documents.py
```diff
@@ -16,7 +16,7 @@ def react_documents(context, module, reload_on_success=False):
     serializer = ChapterSerializer(chapters, many=True)
     chapters_json = JSONRenderer().render(serializer.data).decode("utf-8")
 
-    widget = CKEditorUploadingWidget(config_name="image-editor")
+    widget = CKEditor5Widget(config_name="image-editor")
     widget._set_config()
```
m4ra, line 20: @goapunk this gives error:
```
FAILED tests/documents/dashboard_components/test_views_module_document.py::test_edit_view - 
AttributeError: 'CKEditor5Widget' object has no attribute '_set_config'
```

tests/newsletters/test_newsletter_views.py
```diff
@@ -256,3 +257,67 @@ def test_limit_initiators_organisation_projects(client, project_factory):
...
+    }
+
+    url = reverse(
+        "a4dashboard:newsletter-create", kwargs={"organisation_slug": organisation.slug}
+    )
+    client.login(username=admin.email, password="password")
+    response = client.post(url, data)
+    assert redirect_target(response) == "newsletter-create"
+    assert newsletter_models.Newsletter.objects.count() == 1
+    assert len(mail.outbox) == 1
```
m4ra, line 321: this gives error that is no outbox @goapunk

apps/organisations/forms.py
```diff
@@ -136,7 +135,7 @@ class OrganisationForm(forms.ModelForm):
         ),
         (
             "information",
-            RichTextUploadingFormField,
+            CKEditor5Widget,
             {
                 "config_name": "collapsible-image-editor",
                 "label": OrganisationTranslation._meta.get_field(
```
m4ra, line 141: @goapunk label gives error
```
TypeError: __init__() got an unexpected keyword argument 'label'
```
I guess label is not part of CKEditor5Widget?

adhocracy-plus/config/settings/base.py
```diff
@@ -560,3 +551,134 @@
...
+            "properties": {
+                "styles": "true",
+                "startIndex": "true",
+                "reversed": "true",
+            }
+        },
+        "link": {"defaultProtocol": "https://"},
+    },
+    "image-editor": {
+        "language": "de",
```
goapunk, line 583: we need to keep in mind that we somehow need to fix the language switching, might not have to be in this PR but before we release

apps/documents/assets/ParagraphForm.jsx
```diff
@@ -10,6 +10,21 @@ const ckReplace = function (id, config) {
   return window.CKEDITOR.replace(id, config)
```
goapunk, line 10: all the `ck` functions need to be updated / replaced as well

apps/documents/templates/a4_candy_documents/chapter_detail.html
```diff
@@ -38,11 +38,13 @@ <h2 class="chapter-title">{{ object.name }}</h2>
                         {% for paragraph in object.paragraphs.all %}
 
                             <section class="paragraph" id="paragraph-{{ object.pk }}-{{ paragraph.pk }}">
-                                <div class="paragraph__content">
+                                <div class="paragraph__content ck-content">
```
goapunk, line 41: we only need it on the `paragraph.text` below, not here

apps/organisations/forms.py
goapunk: It's sadly not inheriting from `CharField` I think. We need to set the label in `__init__` I think

apps/topicprio/forms.py
```diff
@@ -1,19 +1,39 @@
...
 
 
-class TopicForm(CategorizableFieldMixin, LabelsAddableFieldMixin, forms.ModelForm):
-
-    description = fields.RichTextUploadingFormField(
-        config_name="image-editor", required=True, label=_("Description")
+class TopicForm(
+    CategorizableFieldMixin,
+    LabelsAddableFieldMixin,
+#    CategoryAndLabelAliasMixin,
```
goapunk, line 16: We don't need this?

docs/ckeditor.md
```diff
@@ -1,30 +1,31 @@
...
+adhocracy4.
 
-## Editor
--   We use [django-ckeditor](https://pypi.org/project/django-ckeditor/) which utilises [ckeditor 4.7](https://ckeditor.com/docs/ckeditor4/latest/index.html).
+# Local Testing
 
-## Configs
--   The configs of the editors can be found in base.py.
--   Config options can be found [here](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html) syntax is slightly different for django-ckeditor (see removeDialogTabs). Official [example config](https://github.com/django-ckeditor/django-ckeditor#example-ckeditor-configuration).
+To test mb with a local version of django-ckeditor-5 you need to follow these
```
goapunk, line 8: mb -> aplus

docs/ckeditor.md
```diff
@@ -1,30 +1,31 @@
...
+npm install
+# build js files
+npm run dev
+```
 
-## Plugins
--   To add plugins, it should be done in a4 in adhocracy4/ckeditor, then update config in aplus (see embedbase), not all ckeditor plugins can be used in django-ckeditor, see below for a list.
--   It is possible to create custom plugins (see collapsibleItem in a4).
--   The embed plugin we use also uses a self hosted version of iframely in order to serve iframes from a urls, configs are found in admin repository.
+Now in mb, do the folowing:
```
goapunk, line 22: mb -> aplus

tests/newsletters/test_newsletter_views.py
goapunk: will investigate, I assume the data we send is wrong, so no email is sent

adhocracy-plus/assets/js/app.js
```diff
@@ -68,11 +68,6 @@ function init () {
     $('.js-select2').select2()
   }
 
-  // This function adds required classes to iframes added by ckeditor
-  $('.rich-text iframe').addClass('ck_embed_iframe')
-  $('.ck_embed_iframe').parent('div').addClass('ck_embed_iframe__container')
-}
```
goapunk, line 74: we need to keep the closing bracket

goapunk (changes requested):
the interactiveevents module needs updating as well and a migration for the iframes, similar to mb

adhocracy-plus/config/settings/base.py
goapunk: upstream django-ckeditor-5 accepted the language PR, so this should be fine now

apps/topicprio/forms.py
```diff
@@ -1,19 +1,39 @@
...
     )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["description"].label = _("Description")
+
     class Meta:
         model = models.Topic
         fields = ["name", "description", "category", "labels"]
+        widgets = {"category": Select2Widget(attrs={"class": "select2__no-search"})}
```
goapunk, line 30: removed this as I think it's mb specific

apps/organisations/migrations/0022_ckeditor_iframes.py
```diff
@@ -0,0 +1,38 @@
...
+from django.utils import translation
+
+
+def replace_iframe_with_figur(apps, schema_editor):
+    template = (
+        '<figure class="media"><div data-oembed-url="{url}"><div><iframe src="'
+        '{url}"></iframe></div></div></figure>'
+    )
+    Organisation = apps.get_model("a4_candy_organisations", "Organisation")
+    translation.activate("en")
```
goapunk, line 14: I wonder if we need to repeat this for all languages, I'm not very familiar with the `parler` package we use for the dynamic translations here

apps/organisations/forms.py
```diff
@@ -175,6 +171,14 @@ def __init__(self, *args, **kwargs):
             for name, field_cls, kwargs in self.translated_fields:
                 self.instance.set_current_language(lang_code)
                 field = field_cls(**kwargs)
+                # The CKEditor5Widget doesn't have a label field, so we need to set it
+                # after creating the object
+                if isinstance(CKEditor5Widget, field):
```
goapunk, line 176: ups, this should be `isinstance(field, CKEditor5Widget)` I think, right? :/

m4ra:
> the interactiveevents module needs updating as well and a migration for the iframes, similar to mb
I don't exactly follow, there is no interactiveevent in mb. can you be more specific?

apps/organisations/forms.py
m4ra: yes i fixed that

goapunk:
> > the interactiveevents module needs updating as well and a migration for the iframes, similar to mb
> 
> I don't exactly follow, there is no interactiveevent in mb. can you be more specific? @goapunk

ah, I think it's called livequestions in mb, but I did that already since I wrote that comment

apps/organisations/forms.py
```diff
@@ -177,7 +162,28 @@ def __init__(self, *args, **kwargs):
         for lang_code in self.languages:
             for name, field_cls, kwargs in self.translated_fields:
                 self.instance.set_current_language(lang_code)
+                print(field_cls)
                 field = field_cls(**kwargs)
+                # The CKEditor5Widget doesn't have a label field, so we need to set it
+                # after creating the object
```
m4ra, line 168: @goapunk I had to bring below also `help_text`, and the plugins fields, or code failed. Now there is still an error message that has to do with django forms:
```
self = <OrganisationForm bound=False, valid=False, fields=(title;logo;image;image_copyright;url;twitter_handle;facebook_handl...tion;nl__slogan;nl__information;ky__description;ky__slogan;ky__information;ru__description;ru__slogan;ru__information)>
name = 'en__information'

    def __getitem__(self, name):
        """Return a BoundField with the given name."""
        try:
            field = self.fields[name]
        except KeyError:
            raise KeyError(
                "Key '%s' not found in '%s'. Choices are: %s." % (
                    name,
                    self.__class__.__name__,
                    ', '.join(sorted(self.fields)),
                )
            )
        if name not in self._bound_fields_cache:
>           self._bound_fields_cache[name] = field.get_bound_field(self, name)
E           AttributeError: 'CKEditor5Widget' object has no attribute 'get_bound_field'
```

apps/organisations/forms.py
```diff
@@ -177,7 +162,28 @@ def __init__(self, *args, **kwargs):
...
+                if isinstance(field, CKEditor5Widget):
```
m4ra, line 169: @goapunk shouldn't this check be against the CKEditor5Field instead of the widget?
ref https://github.com/liqd/django-ckeditor-5/blob/main/django_ckeditor_5/fields.py

apps/organisations/forms.py
```diff
@@ -136,22 +134,9 @@ class OrganisationForm(forms.ModelForm):
         ),
         (
             "information",
-            RichTextUploadingFormField,
+            CKEditor5Widget,
```
goapunk, line 137: this should probabl be `CKEditor5Field`, not widget, sorry

apps/organisations/forms.py
goapunk: oh, yes, and it should be CKEditor5Field above as well I guess

apps/organisations/forms.py
m4ra: tried, didn't work. I try now with charfield and setting the ckeditor5widget as its widget

goapunk: ah, maybe it doesn't play nicely with parler because `CKEditor5Field` doesn't inherit from Charfield

m4ra: @goapunk seems working with charfield

goapunk: after staring at it for a while: I think ` OrganisationTranslation._meta.get_field( "information").formfield` should also work and we don't have to set the widget below then. Maybe worth testing it?

m4ra: Not sure where you want to test the formfield, instead of charfield? the issue with parler compatibility might come back. I leave it as charfield now. You can test it in your PR.

tests/newsletters/test_newsletter_views.py
m4ra: I fxed it

apps/organisations/migrations/0022_ckeditor_iframes.py
m4ra: I don't think we need to set the language here, it takes the default. the translated fields are saved in separate db tables I read in the docs. afaic we would not need to do it for all languages.

apps/organisations/migrations/0023_alter_organisationtranslation_information.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("a4_candy_organisations", "0022_ckeditor_iframes"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="organisationtranslation",
```
m4ra, line 14: hmm, this is a generated PR, and it is fetching the `organisation` model as `organisationtranslation`. Should we do the same in the 0022?
@goapunk

m4ra: I don't see the OrganisationTranslation model, which is also imported in the forms.py from the models.py

goapunk: It's probably created by parler?

apps/organisations/migrations/0022_ckeditor_iframes.py
goapunk: If I remember correctly I added this line because it failed without it.

goapunk: I think we might have to talk to tj about what to do here, it's quite confusing. the ckeditor config allows iframe embeds, however the button isn't actually there to embed them, so in either case we probably don't need this migration because no one was able to add them. We just need to clarify if it should be possible in the new ckeditor

goapunk: @m4ra tj will join standup so we can talk about what to do here

apps/activities/migrations/0002_ckeditor_iframes.py
```diff
@@ -0,0 +1,35 @@
...
+    for activity in Activity.objects.all():
+        soup = BeautifulSoup(activity.description, "html.parser")
+        iframes = soup.findAll("iframe")
+        for iframe in iframes:
+            figure = BeautifulSoup(
+                template.format(url=iframe.attrs["src"]), "html.parser"
+            )
+            iframe.replaceWith(figure)
+        if iframes:
+            activity.live_stream = soup.prettify(formatter="html")
```
goapunk, line 22: this should be `activity.description`, not `live_stream`, forgot to change after copy and pasting

apps/offlineevents/migrations/0004_ckeditor_iframes.py
```diff
@@ -0,0 +1,35 @@
...
+    for offlineEvent in OfflineEvent.objects.all():
+        soup = BeautifulSoup(offlineEvent.description, "html.parser")
+        iframes = soup.findAll("iframe")
+        for iframe in iframes:
+            figure = BeautifulSoup(
+                template.format(url=iframe.attrs["src"]), "html.parser"
+            )
+            iframe.replaceWith(figure)
+        if iframes:
+            offlineEvent.live_stream = soup.prettify(formatter="html")
```
goapunk, line 22: same here :/ should be `description` instead of `live_stream`, sorry

apps/organisations/forms.py
```diff
@@ -177,6 +173,7 @@ def __init__(self, *args, **kwargs):
         for lang_code in self.languages:
             for name, field_cls, kwargs in self.translated_fields:
                 self.instance.set_current_language(lang_code)
+                print(field_cls)
```
goapunk, line 176: forgot to remove ? :)

apps/organisations/migrations/0022_ckeditor_iframes.py
goapunk: we can delete this migration

goapunk (approved):
Really nice!

PR #2578, opened 2024-04-22
fix mixed language on profile page when changing prefered user language there

goapunk:
Testing: go to http://127.0.0.1:8004/account/profile/ and change the language, both alert and rest of the page should be in the new language

**Tasks**
- [ ] Changelog


m4ra (approved):
thanks!

PR #2592, opened 2024-04-24
[8312] comments: refactor and adjust comment styling to new a4 comments

### LINKED ISSUE #2553: Send comment button is too big at brainstroming-module
**URL:** https://aplus-dev.liqd.net
**user:** all user (if logged in)
**expected behaviour:** the button should not be so big
**behaviour:** the button is unusual big
**important screensize:** any
**device & browser:** any
**Comment/Question:** if i delete the "width: 25%;" (as you see in the dev tools in the screenshot) everything works fine.

Screenshot?
Bildschirmfoto 2024-02-01 um 16 22 10

if i deactivate the "width: 25%;": 
Bildschirmfoto 2024-02-01 um 16 21 58

goapunk:
depends on https://github.com/liqd/adhocracy4/pull/1601

there's no design yet with the changes but hopefully it's ok like that

test: on any module which has comments

WIP because it needs the a4 pr merged first and then the a4 hash updated here

also fixes #2553 

**Tasks**
- [x] Changelog


m4ra:
hmm, the button `post` doesn't have a background color. The alert that comment was posted should be on the top, as it is now in the section of the comments, which starts rather low and alert can be missed. I think is best for @hom3mad3 to do the review here and propose some UI solutions. Or have a design from Marcelo.

goapunk:
@m4ra the post button has a background color for me, could you give more info on where it's broken (which module, etc) ? I agree that t he alert isn't ideal but it's also not great on dev either, so maybe we ignore it for now or make an issue as we do need this pr for the release

m4ra:
@goapunk I don't see it broken now with the mapidea and document modules, I should have noted down which module.
There are two little things we can do html wise that can make the big space less and render the success alert more visible.
1. Change the div class from `my-4` to `my-1` in line 159 of the file `apps/contrib/templates/a4_candy_contrib/item_detail.html`
2. Make the header of `Discussion` h3 instead of h2, same as the header `Join the Discussion` in this PR, in [a4 comment box file](https://github.com/liqd/adhocracy4/pull/1593/files#diff-e776a3d4bd21c5b226b381cd5dd1679f54b83a943e7b9ac6ac9d854264b6742bR458) . This will also solve the issue of the different fonts between these headers I see now happening.

m4ra (changes requested):
A minor fix here and in a4 relevant PR, see my comment

goapunk:
@m4ra
for 1. changing `my-4` to `my-1` will also shrink the margin at the bottom, not sure if we want this?
for 2. I'm unsure if changing to h3 is a good idea accessibility wise, I think technically both of them should actually be h2 instead of h3 to create a logical structure?

At this point we should probably ask @mcastro-lqd how to best proceed then. @mcastro-lqd The problem for the new comments is that the alert is quite further down when posting a new comment via the "Join the discussion" form (see screenshot below). I'm also now unsure if the headings are correct (see the screenshot as well). Any suggestion what we should do here?

Screenshot 2024-05-21 at 14-39-07 adhocracy _liqd-orga

mcastro-lqd:
Would "using a CSS Framework utilizing utility classes to adjust the heading size. For example, with Bootstrap, you can use classes like h3, h4, etc., to adjust the heading size while maintaining semantic <h2> tags." be the case? Cause otherwise it's too big.

goapunk:
@m4ra @mcastro-lqd I changed both to be h3 now, so it's a bit smaller

goapunk:
and decreased the margin-top for the post button as it was much bigger than on dev

m4ra (approved):
thanks!

goapunk:
@m4ra opened https://github.com/liqd/adhocracy4/pull/1601 so this can be reviewed again

Issue #2597, opened 2024-04-29
project page: organisation logo bugged 


TeaJayyy:
**URL:** https://aplus-dev.liqd.net/permanent-test-project/projects/umfrage-6/
**user:** any
**expected behaviour:** 
**behaviour:** see screenshot
**important screensize:** any
**device & browser:** any
**Comment/Question:**

<img width="1451" alt="Bildschirmfoto 2024-04-29 um 13 35 50" src="https://github.com/liqd/adhocracy-plus/assets/147062890/a54e4a58-5147-42b3-afe5-16967c93a6ce">

PR #2598, opened 2024-04-29
scss/components: fix .tablist and organization logo z-index

hom3mad3:
fixes https://github.com/liqd/adhocracy-plus/issues/2597


goapunk (approved):
nice :)

PR #2601, opened 2024-04-30
[8103] apps/cms: add a new ImportantPage "registration" to wagtail.

goapunk:
This new page can be used to create an explanation page about why registration is necessary. If the page is set in wagtail, there will now be link on the signup page saying "Why register?".

Testing: in wagtail create a new page and then create a new child page of that page (simplepage). After saving, go to settings -> important pages and set the newly created simplepage for registration. There should now be a new link on the signup page

**Tasks**
- [x] Changelog


adhocracy-plus/templates/account/signup.html
```diff
@@ -6,6 +6,10 @@
 {% block content %}
     <h1>{% translate "Register" %}</h1>
 
+    {% if settings.a4_candy_cms_settings.ImportantPages.registration.url %}
+        <a href="{{ settings.a4_candy_cms_settings.ImportantPages.registration.url }}" target="_blank">{% blocktranslate %}Why register?{% endblocktranslate %}</a>
+    {% endif %}
+
```
m4ra, line 12: It needs an extra check for the page to be published and not in draft mode. As it is now the link appears even if  the register page is a draft

goapunk: hmm, I guess that raises the question if we need this check at all, we don't have a check for the other pages (e.g. in the checkbox for the data protection page, or the footer), I just thought it's nice to hide it. The story doesn't say it should only be shown if the page exists. I will look into how easy it is to check if the page is draft

goapunk: @m4ra added a check

m4ra: thanks! can we have a nice wrap of this long variable of the important page using the [with](https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#with) block. and a commit squash?

goapunk: done

m4ra (approved):
Nice, thank u!

PR #2607, opened 2024-05-15
[8152] Add anchor for ordering

m4ra:
**Tasks**
- [x] Changelog


m4ra:
@goapunk @hom3mad3 I used the existing `index` id from a previous story as an anchor, and added this id in the `ordering options`.
image

 I cannot tweak the search button though, as when we (only) search something, the URL is constructed with search and the ordering parameter is added by default, because we have set the ordering filter of most recent as the default one in [views](https://github.com/liqd/adhocracy-plus/blob/be4b64c54400cfd5b97627dfe64c88c58c2032b2/apps/ideas/views.py#L41). And I cannot find how to customize this.
image

goapunk (approved):
LGTM

PR #2618, opened 2024-05-23
fix #2585 a4_candy_documents/chapter_detail: expand toc on all chapters

### LINKED ISSUE #2585: #7637 [a+] big django & wagtail update: Inconsistency in "Text Discussion" chapter overview
**URL:**
https://aplus-dev.liqd.net/permanent-test-project/projects/text-diskutieren-einzelmodul/

**user:**
Any

**expected behaviour:**
When navigating through the chapters using the table of contents menu, I expect the state of the menu (collapsed or opened) to remain the same, when switching to another chapter.

**behaviour:**
When navigating to the first chapter, the menu always opens. When navigating to any subsequent chapter, the menu is collapsed by default. Wether or not I collapse or open the menu by myself before switching, doesn't influence the bahaviour of the menu.

**important screensize:**
Any

**device & browser:**
Current MacOS, Firefox 124.0.2

hom3mad3:
fixes https://github.com/liqd/adhocracy-plus/issues/2585

### DIFF
--- a/apps/documents/templates/a4_candy_documents/chapter_detail.html
+++ b/apps/documents/templates/a4_candy_documents/chapter_detail.html
@@ -10,15 +10,15 @@
                     {% if chapter_list|length > 1 %}
                         <nav class="doc-toc" aria-label="{% translate 'Table of Contents' %}">
                             <button
-                                class="doc-toc__toggle {% if object.prev %}collapsed{% endif %}"
+                                class="doc-toc__toggle"
                                 data-bs-toggle="collapse"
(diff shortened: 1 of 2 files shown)


goapunk:
@m4ra @hom3mad3 I think the always open by default version in this PR makes more sense but eventually that's for @TeaJayyy to decide I'd say

goapunk (changes requested):
LGTM but should have a changelog as it changes existing behavior

goapunk (approved):
Nice!

PR #2629, opened 2024-05-27
[8102] remove unique email to enable social account autoconnect 

### LINKED ISSUE #2606: #8102 [Z:T] Enable more SSOs on a+: Issue with Duplicate Registration Email and Confirmation Handling
**Expected Behavior:** 
When a user attempts to register with an email address already associated with an existing account (either social or normal email account), they should:

1. Receive an email reminding them of their existing account and providing instructions for password reset.
2. See the standard "we have sent you a confirmation link" message on the website to prevent unauthorized users from determining if an email is registered.

**Current Behavior:**
When registering with a social account, email confirmation is not mandatory. This can cause confusion if the user forgets they used social login and tries to register again using the same email, resulting in an error and preventing successful registration/ handling of the duplicate accounts.

**Steps to Reproduce**
1. Register on the website using a social account (e.g., Google, Facebook) without confirming the email.
3. Log out.
4. Attempt to register again using the same email address (either through a social account or normal email registration).

**Screenshots**
Error message after attempting to register with an e-mail adress that is already linked to a social account but hasn't been confirmed in that account.
Bildschirmfoto 2024-05-15 um 09 44 36

m4ra:
fixes #2606 

**Tasks**
- [x] Changelog


adhocracy-plus/config/settings/base.py
```diff
@@ -283,17 +285,18 @@
...
 ACCOUNT_LOGIN_ON_PASSWORD_RESET = True
 ACCOUNT_USERNAME_MIN_LENGTH = 5
 ACCOUNT_USERNAME_REQUIRED = True
-ACCOUNT_PREVENT_ENUMERATION = True
+ACCOUNT_PREVENT_ENUMERATION = "strict"
 SOCIALACCOUNT_AUTO_SIGNUP = False
 SOCIALACCOUNT_EMAIL_VERIFICATION = "none"
 SOCIALACCOUNT_FORMS = {"signup": "apps.users.forms.SocialTermsSignupForm"}
 SOCIALACCOUNT_QUERY_EMAIL = True
+SOCIALACCOUNT_EMAIL_AUTHENTICATION = True
```
goapunk, line 298: @m4ra from my testing and research this apparently will only work with providers which verify the user email (which makes sense security-wise). From what I could find out in our case that only is apple and google. For github and facebook it won't work, so I wonder if it makes sense to enable this option at all. Not sure how to proceed here, what do you think?

m4ra: I changed the socialaccount verification to mandatory, so if not verified from provider, the email verification should be sent (this is what I understand from the allauth documentation). Can you pull and have another try? The dummy provider doesn't work in this scenario. It gives error that the field is required, aka the email field, which is a hidden form and from the browser inspect it's value is a long string, I guess it's encrypted?
image

apps/users/models.py
```diff
@@ -39,8 +39,7 @@ class User(auth_models.AbstractBaseUser, auth_models.PermissionsMixin):
 
     email = models.EmailField(
         _("Email address"),
-        unique=True,
-        error_messages={"unique": _("Email is invalid or already taken.")},
+        blank=True,
```
goapunk, line 42: is there any case where we would want the email to be blank?

apps/users/adapters.py
```diff
@@ -54,13 +54,13 @@ def get_email_confirmation_url(self, request, emailconfirmation):
...
-            request.GET["next"], allowed_hosts=None
+    def get_email_verification_redirect_url(self, email_address):
+        if "next" in self.request.GET and url_has_allowed_host_and_scheme(
+            self.request.GET["next"], allowed_hosts=None
         ):
-            return request.GET["next"]
+            return self.request.GET["next"]
         else:
-            return super().get_email_confirmation_redirect_url(request)
+            return super().get_email_verification_redirect_url(email_address)
```
m4ra, line 63: @goapunk this was changed in allauth codebase. Don't know why email_address is there, as I don't see it being used in the super's method either, but this is how it's constructed.

adhocracy-plus/config/settings/base.py
```diff
@@ -37,6 +37,7 @@
     "allauth",
     "allauth.account",
     "allauth.socialaccount",
+    "allauth.socialaccount.providers.dummy",
```
m4ra, line 40: moving these to local settings @goapunk

adhocracy-plus/config/settings/dev.py
```diff
@@ -14,7 +14,10 @@
 except ImportError:
     pass
 else:
-    INSTALLED_APPS += ("debug_toolbar",)
+    INSTALLED_APPS += (
+        "allauth.socialaccount.providers.dummy",
```
goapunk, line 18: this ties the dummy provider to whether the debug_toolbar is installed or not, wouldn't it be a bit clearer to have a separate `INSTALLED_APPS += (""allauth.socialaccount.providers.dummy")` somehwere outside the try block?

adhocracy-plus/config/settings/dev.py
```diff
@@ -9,6 +9,8 @@
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = "k3l^@x53$l5@y(fo6ivgplj&q^^rmtl^enzse5ozxuepx0$s()"
 
+INSTALLED_APPS += ("allauth.socialaccount.providers.dummy",)
+
```
m4ra, line 13: good catch @goapunk

apps/users/models.py
```diff
@@ -127,8 +125,7 @@ class User(auth_models.AbstractBaseUser, auth_models.PermissionsMixin):
 
     objects = auth_models.UserManager()
 
-    USERNAME_FIELD = "email"
-    REQUIRED_FIELDS = ["username"]
```
goapunk, line 131: Do we want "email" as required field here?

requirements/base.txt
```diff
@@ -16,7 +16,7 @@ redis==5.0.4
 
 # Inherited a4-core requirements
 Django==4.2.13
-django-allauth==0.55.0
+django-allauth==0.63.2
```
goapunk, line 19: should this be `django-allauth[socialaccount]` as described here https://docs.allauth.org/en/latest/release-notes/recent.html#backwards-incompatible-changes ?

goapunk (commented):
Really nice! Just some small questions

goapunk (approved):
Really cool, thanks!

Issue #2599, opened 2024-04-30
moderator dashboard: organisation list overlaps with comments / reports on medium screens


goapunk:
**URL:** https://aplus-dev.liqd.net/userdashboard/moderation/detail/jonathan-app-test/
**user:** moderator / admin
**expected behaviour:** organisation names and comment tiles don't overlap
**behaviour:** organisation names and comment tiles overlap on medium screens
**important screensize:** 560 < x < 760
**device & browser:** 
**Comment/Question:** 

Screenshot?
![Screenshot 2024-04-30 at 10-38-27 adhocracy](https://github.com/liqd/adhocracy-plus/assets/1710203/befee0dc-927e-4ee7-9d1c-246a4372f6bc)

PR #2630, opened 2024-05-27
templates/base_userdashboard: fix text overlap

hom3mad3:
fixes https://github.com/liqd/adhocracy-plus/issues/2599

### DIFF
--- a/apps/userdashboard/templates/a4_candy_userdashboard/base_userdashboard.html
+++ b/apps/userdashboard/templates/a4_candy_userdashboard/base_userdashboard.html
@@ -89,7 +89,7 @@
                                 aria-label="{% if organisation.title %}{{ organisation.title }}{% else %}{{ organisation.name}}{% endif %} {% trans 'Home' %}"
                               ></i>
                             {% endif %}
-                            <div class="ms-3 text-start">{% if organisation.title %}{{ organisation.title }}{% else %}{{ organisation.name}}{% endif %}</div>
+                            <div class="ms-3 text-start u-word-break">{% if organisation.title %}{{ organisation.title }}{% else %}{{ organisation.name}}{% endif %}</div>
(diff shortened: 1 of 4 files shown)


apps/userdashboard/templates/a4_candy_userdashboard/base_userdashboard.html
```diff
@@ -89,7 +89,7 @@ <h2 class="fs-6">{% trans 'Initiator Organisations' %}</h2>
                                 aria-label="{% if organisation.title %}{{ organisation.title }}{% else %}{{ organisation.name}}{% endif %} {% trans 'Home' %}"
                               ></i>
                             {% endif %}
-                            <div class="ms-3 text-start">{% if organisation.title %}{{ organisation.title }}{% else %}{{ organisation.name}}{% endif %}</div>
+                            <div class="ms-3 text-start u-word-break-all">{% if organisation.title %}{{ organisation.title }}{% else %}{{ organisation.name}}{% endif %}</div>
```
goapunk, line 92: could we also use `u-break-word` instead of `u-word-break-all` as the latter will always break it unnaturally whereas `u-break-word` at least tries to brake the word at a natural point (or move it to a new line)

goapunk (commented):
@hom3mad3 generally works but a question about the unnatural word breaking

apps/userdashboard/templates/a4_candy_userdashboard/base_userdashboard.html
goapunk: I might misunderstand how `u-break-word` works, from what I understand it will:
1. try to break the word naturally
2. move it to a newline if 1 doesn't work
3. break it unnaturally if 1 and 2 fail

That should still prevent the overflow then?

PR #2707, opened 2024-06-13
adhocracy-plus/templates: fix outdated phone number, fix redirects to

goapunk:
root instead of hardcoded value

- fix template lint errors and reformat

fixes https://github.com/liqd/a4-defakts/issues/46

**Tasks**
- [x] Changelog


adhocracy-plus/templates/500.html
```diff
@@ -16,48 +17,58 @@
...
-                    <p><a href="https://liqd.net">Liquid Democracy e.V.</a> Am Sudhaus 2, 12053, Berlin, tel. +49 (0) 30 6298 4840, e-mail. hallo@liqd.net</p>
+            </div>
+        </header>
+        <main id="main">
+            <div class="small-wrapper">
+                <div class="error_body">
+                    <h1>500</h1>
+                    <p>Internal server error.</p>
+                    {# djlint:off D018 #}
+                    <a href="/">Back to adhocracy+</a>
```
goapunk, line 60: I assume the idea in the error template was to not use any django functionality in case they throw an error, so I added the djlint exception here instead of using the `{% url %}` tag as suggested by djlint

m4ra (approved):

PR #2713, opened 2024-06-18
apps/projects: make insights migration a bit faster by using select_r…

goapunk:
…elated

**Tasks**
- [ ] Changelog

### DIFF
--- a/apps/projects/migrations/0006_initialize_insights.py
+++ b/apps/projects/migrations/0006_initialize_insights.py
@@ -21,34 +21,44 @@ def initialize_insights(apps, schema_editor):
     user_ids = defaultdict(set)
 
     for idea_model in [Idea, MapIdea, Proposal]:
-        for idea_object in idea_model.objects.all():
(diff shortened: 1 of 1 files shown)


apps/projects/migrations/0006_initialize_insights.py
```diff
@@ -21,34 +21,46 @@ def initialize_insights(apps, schema_editor):
...
+    for comment in Comment.objects.exclude(project=None).select_related(
+        "project", "creator"
+    ):
         project = comment.project
         insights[project].comments += 1
         user_ids[project].add(comment.creator.id)
 
-    for like in Like.objects.all():
+    for like in Like.objects.all().select_related(
+        "livequestion", "livequestion__module", "livequestion__module__project"
```
m4ra, line 58: only need to fetch the "livequestion__module__project" iirc @goapunk

goapunk: ah, yes, forgot that

m4ra (approved):
good, feel free to push to main

PR #2715, opened 2024-06-19
apps/users: use magic as a fallback to detect image mimetype if MIMEI…

goapunk:
…mage fails

see https://sentry.liqd.net/organizations/liqd/issues/3774

**Tasks**
- [ ] Changelog


apps/users/emails.py
```diff
@@ -47,7 +48,12 @@ def get_attachments(self):
         organisation = self.get_organisation()
         if organisation and organisation.logo:
             f = open(organisation.logo.path, "rb")
-            logo = MIMEImage(f.read())
```
m4ra, line 50: just assign the logo to None here, as python doesn't like calling variables not set outside conditional statements (we saw this bug in allauth)

apps/users/emails.py
```diff
@@ -47,7 +48,12 @@ def get_attachments(self):
...
+            try:
+                logo = MIMEImage(f.read())
+            except TypeError:
+                f.seek(0)
+                filetype = magic.from_buffer(f.read(), mime=True)
+                logo = MIMEImage(f.read(), _subtype=filetype)
```
m4ra, line 56: here I would check `if logo:` and then add it to the attachements, else I would send a notification/log error. And IMO attachements should be returned eitherway, even if the logo didn't make it, otherwise emails gets lost.

m4ra (changes requested):
Don't forget to add the package in the requirements (and the server needs the libmagic lib) and some small code additions

apps/users/emails.py
m4ra: maybe better to use magic in the try block, and in the exception we send an email error and log?

goapunk: I just checked and magic doesn't throw an exception, if it's garbage data it will return `application/octet-stream` as mime type

goapunk: but I can move it in it's on try if you want

apps/users/emails.py
```diff
@@ -47,9 +49,23 @@ def get_attachments(self):
...
         if organisation and organisation.logo:
             f = open(organisation.logo.path, "rb")
-            logo = MIMEImage(f.read())
-            logo.add_header("Content-ID", "<{}>".format("organisation_logo"))
-            attachments += [logo]
+            logo = None
+            try:
+                logo = MIMEImage(f.read())
+            except TypeError:
+                capture_message(
```
goapunk, line 56: @m4ra added a sentry notification here

m4ra (approved):
thanks|

apps/users/emails.py
```diff
@@ -46,10 +48,24 @@ def get_attachments(self):
...
+                    capture_message(
+                        "warning: MIMEImage failed to detect mime type:\n"
+                        "organisation:"
+                        + organisation.name
+                        + "\nfile:"
+                        + organisation.logo.path
+                    )
+                    mime_type = magic.from_buffer(data, mime=True)
+                    logo = MIMEImage(data, _subtype=mime_type)
+                if logo:
```
m4ra, line 66: I'd decouple the `with` block from checking if we have a logo, and thus assign the logo variable also outside the `with`,

m4ra: Better decouple the logo checking from the `with` block, and so the first assignment of the logo set just before the `with`.

m4ra (approved):

PR #2724, opened 2024-06-26
apps/users: make switching the language via the indicator also change

### LINKED ISSUE #2719: toggle language bugged with a set preferred languages
**URL:** https://adhocracy.plus/musterorganisation/ideas/2021-04618
**user:** any 
**expected behaviour:** the comment section is displayed in the language I chose 
**behaviour:** the comment section is always in German 
**important screensize:** any
**device & browser:** any

Notes: Having a set prefered language in the account settings will lead to some weird, partially translated pages when changing the language toggle. As you can see in the screenshot below, I changed the language to English but my prefered language is German. So the comment section stays in German but some other elements like date, refrence no. or "go back to overview" are translated in English.
 
I prefer, although it's nice to have a globally prefered language, that the language toogle on top works as expected and therefore translates every element on the page.

grafik

goapunk:
the users preferred language.

- add missing cookie settings to language cookie

fixes #2719

**Tasks**
- [x] Changelog

### DIFF
--- a/apps/users/middleware.py
+++ b/apps/users/middleware.py
@@ -13,6 +13,13 @@ class SetUserLanguageCookieMiddleware:
             translation.activate(request.user.language)
             if cookie != request.user.language:
                 response.set_cookie(
-                    settings.LANGUAGE_COOKIE_NAME, request.user.language
+                    settings.LANGUAGE_COOKIE_NAME,
+                    request.user.language,
+                    max_age=settings.LANGUAGE_COOKIE_AGE,
+                    path=settings.LANGUAGE_COOKIE_PATH,
+                    domain=settings.LANGUAGE_COOKIE_DOMAIN,
+                    secure=settings.LANGUAGE_COOKIE_SECURE,
+                    httponly=settings.LANGUAGE_COOKIE_HTTPONLY,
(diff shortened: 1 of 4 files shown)


apps/users/middleware.py
```diff
@@ -13,6 +13,13 @@ def __call__(self, request):
             translation.activate(request.user.language)
             if cookie != request.user.language:
                 response.set_cookie(
-                    settings.LANGUAGE_COOKIE_NAME, request.user.language
+                    settings.LANGUAGE_COOKIE_NAME,
```
goapunk, line 16: taken from https://github.com/django/django/blob/main/django/views/i18n.py#L64C12-L73C14

m4ra:
@goapunk this fix has a side-effect that when we navigate the website and at some point switch to a language different from our preferences, when we navigate back to our account settings, then this new language becomes as our preferred (the language input becomes this chosen language we toggled). While currently whenever we navigate back to our account settings, it switches back to our default/preferred language. Is that ok @TeaJayyy?

m4ra:
@goapunk did a quick demo to Tietje and the preferred language change is what should be expected, so am going to merge this.

m4ra (approved):
thank you :)

PR #2737, opened 2024-08-01
Change the wording of the help text for the map component

P4rcev4l:
not sure how things work with translation


P4rcev4l:
I'm not sure if I changed the correct files, in fact I'm pretty sure I didn't given that the CI fail. I need some pointers here

goapunk:
@P4rcev4l I think the correct place to change it is here: https://github.com/liqd/adhocracy-plus/blob/main/apps/mapideas/models.py#L16

goapunk:
added the missing migrations

goapunk (approved):
LGTM, thanks!

PR #2748, opened 2024-08-19
[7619] image deletion docs

m4ra:
**Tasks**
- [x] Changelog

### DIFF
--- a/apps/debate/views.py
+++ b/apps/debate/views.py
@@ -96,7 +96,7 @@ class SubjectUpdateView(
 class SubjectDeleteView(
     mixins.DashboardBaseMixin,
     mixins.DashboardComponentMixin,
-    mixins.DashboardComponentDeleteSignalMixin,
+    mixins.DashboardComponentFormSignalMixin,
     idea_views.AbstractIdeaDeleteView,
 ):
     model = models.Subject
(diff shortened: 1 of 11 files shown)


apps/debate/views.py
```diff
@@ -96,7 +96,7 @@ def get_permission_object(self):
 class SubjectDeleteView(
     mixins.DashboardBaseMixin,
     mixins.DashboardComponentMixin,
-    mixins.DashboardComponentDeleteSignalMixin,
+    mixins.DashboardComponentFormSignalMixin,
```
m4ra, line 99: @goapunk this has been deprecated for a while. and we had left a trial period when we switched to `form_valid` for deletion.

m4ra:
@goapunk will add few more tests for the modules in a separate commit, but you can start reviewing it.

docs/deletion.md
```diff
@@ -0,0 +1,90 @@
+# Deletion of images conforming to DSGVO
```
goapunk, line 1: For all of these cases: should we add a sentence explaining that an old image gets deleted when it's a replaced by a new one (and not just when e.g. an idea is deleted) ?

goapunk (commented):
Small comment about image replacement and should we also add tests which checks that the old image is deleted when uploading a new one? (can be in a separate PR)

docs/deletion.md
goapunk: also one of the tests is failing :)

m4ra:
@goapunk there is a test for users replacing images, and I added for project/organisation and interactiveevent. i won't add for the other modules as should be same behaviour as the interactiveevent.

PR #2764, opened 2024-09-17
[8304] Add geolocation tests

### LINKED ISSUE #2767: API showing only app projects
change code, so that all projects on a+ are shown with the API

m4ra:
Fixes  #2767

**Tasks**
- [x] Changelog

### DIFF
--- a/apps/projects/api.py
+++ b/apps/projects/api.py
@@ -27,6 +27,7 @@ class AppProjectsViewSet(viewsets.ReadOnlyModelViewSet):
             module__phase__end_date__gt=now,
             is_draft=False,
             is_archived=False,
+            organisation__enable_geolocation=True,  # TODO: replace with a django filter later
         )
 
 
--- a/apps/projects/api.py
+++ b/apps/projects/api.py
@@ -1,7 +1,10 @@
+from django.db.models import Q
+from django.utils import timezone
 from rest_framework import permissions
 from rest_framework import viewsets
 
 from adhocracy4.modules.models import Module
+from adhocracy4.projects.enums import Access
 from adhocracy4.projects.models import Project
 
 from .serializers import AppModuleSerializer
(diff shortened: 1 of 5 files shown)


tests/projects/test_project_geolocation.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # Create a Project instance with the GeoJSON point
+    project.point = geojson_point
+    project.save()
+
+    fetched_project = Project.objects.get(id=project.id)
+
+    # Check if the point is correctly stored
+    assert fetched_project.point == geojson_point
+    print(fetched_project.point)
```
goapunk, line 27: do we need this ? :)

tests/projects/test_project_geolocation.py
```diff
@@ -0,0 +1,49 @@
...
+    ]  # Check coordinates
+
+
+@pytest.mark.django_db
+def test_project_geojson_point_serialiser(
+    user, project_factory, module_factory, phase_factory, apiclient
+):
+    project = project_factory(is_app_accessible=True, point=geojson_point)
+    module = module_factory(project=project)
+    #    module_factory(project=project, is_draft=True)
```
goapunk, line 40: do we need this ? :)

goapunk (changes requested):
Nice, just a small note on some commented code / print.

m4ra:
@goapunk added also the small fix for the app projects filter

apps/projects/api.py
```diff
@@ -15,9 +15,7 @@ class AppProjectsViewSet(viewsets.ReadOnlyModelViewSet):
     lookup_field = "slug"
 
     def get_queryset(self):
-        return Project.objects.filter(
-            is_draft=False, is_archived=False, is_app_accessible=True
-        )
+        return Project.objects.filter(is_draft=False, is_archived=False)
```
goapunk, line 18: looking at this, I wonder if this will show private projects? if so we definitely don't want to do that, maybe we can add a test for it to make sure they are not included ? this will also answer the question then

goapunk: on a side note: maybe we can add a fixme/todo in line 12 to rename it from `AppProjectsViewSet` to `ProjectViewSet` as this is no longer app specific

m4ra: And what about past projects? I think @TeaJayyy needs to step in and tell us if the API should only show current and future public projects, so no private or semi-private.

apps/projects/api.py
```diff
@@ -1,22 +1,27 @@
...
+# FIXME:rename it from AppProjectsViewSet to ProjectViewSet
 class AppProjectsViewSet(viewsets.ReadOnlyModelViewSet):
     serializer_class = AppProjectSerializer
     permission_classes = (permissions.IsAuthenticated,)
     lookup_field = "slug"
 
     def get_queryset(self):
         return Project.objects.filter(
-            is_draft=False, is_archived=False, is_app_accessible=True
+            Q(access=Access.PUBLIC) | Q(access=Access.SEMIPUBLIC),
```
goapunk, line 22: according to @TeaJayyy comment the api should only show current and future projects, this will also include past projects, right?

goapunk (commented):
Looks good to me, only a question about the past projects

apps/projects/api.py
m4ra: good point, added the extra filter and a relevant test

goapunk (approved):
nice!

goapunk (changes requested):
tests are now failing

goapunk (approved):
Nice, looks good to me!

PR #2790, opened 2024-09-30
Changed datepicker for flatpicker

rjlanari:
*Replace the old datepicker with flatpickr and connect the flatpickr instances (if there are multiple ones, e.g. for module phases).
- [x] Changelog


requirements/dev.txt
```diff
@@ -10,7 +10,6 @@ freezegun==1.5.1
 isort==5.13.2
 mkdocs==1.6.1
 mkdocstrings[python]==0.26.1
-psycopg[binary]==3.2.2
```
m4ra, line 13: you need to add this back. I guess was deleted by accident. If you read the tests failing you can see that the psycopg library is missing.

m4ra (changes requested):

requirements/dev.txt
rjlanari: Thanks @m4ra! I erased that line because there was some versions problems with it! I upgraded the OS and now I'm installing Xcode dev tools again! I will add the line and let you know!

package.json
```diff
@@ -25,6 +25,7 @@
     "datepicker": "git+https://github.com/liqd/datePicker.git",
```
m4ra, line 25: shouldn't we remove this? @goapunk I dunno why in mB is still there.

m4ra:
@rjlanari @goapunk the flatpicker in mB works as dropdown, while the one here is similar to datepicker.
**from mB dev:**
image

**from this a+ PR:**
image

rjlanari:
> @rjlanari @goapunk the flatpicker in mB works as dropdown, while the one here is similar to datepicker. **from mB dev:** image
> 
> **from this a+ PR:** image

Hello @mara, why can it be that when I test it locally I get it as dropdown?


Screenshot 2024-10-02 at 11 17 23

m4ra:
@rjlanari you are right! I didn't rebuild the npm. Works as expected now, I will go ahead and merge. Thank you

m4ra (approved):
Nice!

m4ra:
@rjlanari can you run a rebase with main from your branch?
```
git fetch
git checkout your-branch
git rebase main
```

package.json
goapunk: yes, we don't need this anymore. I probably forgot to remove it

adhocracy-plus/assets/js/init-picker.js
```diff
@@ -1,15 +1,77 @@
...
+
+  // link non-phase datepickers if exist
+  const fStart = flatpickrs.get(idStart)
+  const fEnd = flatpickrs.get(idEnd)
+  if (fStart && fEnd) {
+    link(fStart, fEnd)
+  }
+
+  // link phase datepickers if exist
+  for (let i = 0; i < phaseIds.length - 1; i++) {
```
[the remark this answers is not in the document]
goapunk: I think that's on purpose (but can't recall). We always link phase[i] with phase[i+1] (see line 43). If `i` went up to the last element it would try to link it with last element + 1 which doesn't exist

adhocracy-plus/assets/js/init-picker.js
```diff
@@ -1,15 +1,77 @@
...
+    const p0 = flatpickrs.get(phaseIds[i])
+    const p1 = flatpickrs.get(phaseIds[i + 1])
+    if (p0 && p1) {
+      link(p0, p1)
+    }
+  }
+}
 
-  $inputs.addClass('form-control')
+function initDatePicker () {
```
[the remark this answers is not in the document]
goapunk: not sure if that's something @rjlanari wants to take over as the focus was more on django, what do yo u think @rjlanari ? Otherwise maybe you can work on that @hom3mad3 if you have an idea in mind? If we change it we should port it over to mb (or if both files are identical we could even consider moving it into a4)

PR #2830, opened 2024-11-04
[8381] add unregistered participants field to project insights to account fo…

goapunk:
…r the

new feature in the poll module which allows participating without account

**Tasks**
- [x] Changelog

depends on https://github.com/liqd/adhocracy4/pull/1688


apps/projects/signals.py
```diff
@@ -93,5 +93,16 @@ def increase_poll_answers_count(sender, instance, created, **kwargs):
 
         insight, _ = ProjectInsight.objects.get_or_create(project=project)
         insight.poll_answers += 1
-        insight.active_participants.add(instance.creator.id)
+        if instance.creator:
+            insight.active_participants.add(instance.creator.id)
+        else:
+            answers = Answer.objects.filter(
```
goapunk, line 99: The problem with the current way of counting participants is that it is tied to the post_save signal of Vote and Answer. As a user can have multiple Votes + Answers in a single Poll, we can't just increase `insight.unregistered_participants` by one every time this signal triggers but have to check if there's a vote/answer with the same content_id already, and if there's none (or only one as we don't know if the answer/vote was already saved to the db when `post_save` is called). Not sure if this solution is any good, signals should not be long running and also not sure if this is safe, post_save doesn't guarantee that the model has been written to the database, so this might be slightly inaccurate in some edge cases ? we could wrap it in `transaction.on_commit` but this creates other problems, so I'm not really sure what the  best way forward is. Some ideas could be:
* ignore unregistered users in insights (not great)
* do like in this PR and live with possible inaccuracy
* have an extra model which has project id and content _id to check if the content_id has already participated in the project
* find another way to track poll participants (instead of signal on  Vote and Answer model), probably tricky because insights are in a+ and poll is in a4, but we could probably have a signal in a4 which we connect to a signal here

apps/projects/signals.py
```diff
@@ -93,5 +93,16 @@ def increase_poll_answers_count(sender, instance, created, **kwargs):
 
         insight, _ = ProjectInsight.objects.get_or_create(project=project)
         insight.poll_answers += 1
-        insight.active_participants.add(instance.creator.id)
+        if instance.creator:
+            insight.active_participants.add(instance.creator.id)
```
m4ra, line 97: I would include only this check for now. The wider question of whether to include unregistered participants in the insights is not part of this story scope. It needs to be discussed with @TeaJayyy in a later stage. So I suggest we leave this PR as a draft. And you open a new PR with a small fix.

apps/projects/signals.py
m4ra: I think I prefer the intermediary model with project and content relations. We can talk about it in a future story.

m4ra (commented):
I suggest to keep this PR on-hold and proceed with the fix in a new PR. See my inline comments.

apps/projects/signals.py
TeaJayyy: @m4ra As discussed with @goapunk, we need an answer (or small fix) for that issue now, before we release the new poll. The solution with some inaccuracies but still a somewhat valid representation of users that participated should be fairly doable for now, right? 

I agree - we can implement a more sophisticated solution later.

goapunk:
I've updated this with the signal solution to compare now

m4ra:
@goapunk there is an error also with the signals import, tests failing

apps/projects/models.py
```diff
@@ -134,7 +135,10 @@ def create_insight_context(insight: ProjectInsight) -> dict:
     show_ideas = bool(blueprint_types.intersection({"BS", "IC", "MBS", "MIC", "PB"}))
 
     counts = [
-        (_("active participants"), insight.active_participants.count()),
+        (
+            _("active participants"),
+            insight.active_participants.count() + insight.unregistered_participants,
```
m4ra, line 140: here active participants are the sum of both registered and unregistered. But when a poll is submitted by an unregistered user, the active participants are not increased, only the unregistered ones. See my comment below in the signal.

apps/projects/signals.py
```diff
@@ -93,6 +94,14 @@ def increase_poll_answers_count(sender, instance, created, **kwargs):
...
         insight.save()
+
+
+@receiver(poll_voted)
+def increase_poll_participant_count(sender, poll, creator, content_id, **kwargs):
+    insight, _ = ProjectInsight.objects.get_or_create(project=poll.module.project)
+    if creator:
+        insight.active_participants.add(creator.id)
+    else:
+        insight.unregistered_participants += 1
```
m4ra, line 106: here, shouldn't we also increment the `active_participants`?

m4ra (changes requested):
Some clarifications on how we count active participants

apps/projects/signals.py
```diff
@@ -93,6 +94,14 @@ def increase_poll_answers_count(sender, instance, created, **kwargs):
...
         insight, _ = ProjectInsight.objects.get_or_create(project=project)
         insight.poll_answers += 1
-        if instance.creator:
-            insight.active_participants.add(instance.creator.id)
         insight.save()
+
+
+@receiver(poll_voted)
+def increase_poll_participant_count(sender, poll, creator, content_id, **kwargs):
+    insight, _ = ProjectInsight.objects.get_or_create(project=poll.module.project)
```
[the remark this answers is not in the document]
goapunk: sorry, what do you mean ? we create a `ProjectInsight` if it doesn't exist already for the project as the `ProjectInsight` isn't created automatically for a project, right?

apps/projects/models.py
goapunk: `active_participants` is a m2m relation to the user model, so we can't increase it for unregistered users, that's why I added the `unregistered_participants` field. And we combine both fields here to reflect the total count of participants (maybe somewhat confusingly still called "active participants").

m4ra (approved):
thanx for clarifying, LGTM

m4ra:
@goapunk can you update the a4 hash, and add a changelog? I have approved already, so please go ahead and merge once you are done.

m4ra:
@goapunk looks like the project-insight model needs to have the active_participants field as nullable. See the test failing.

apps/projects/insights.py
```diff
@@ -68,16 +68,41 @@ def create_insight(project: Project) -> ProjectInsight:
...
     for obj in creator_objects:
-        ids = list(obj.values_list("creator", flat=True).distinct().order_by())
+        # ignore objects which don't have a creator, they are counted in the next step.
+        ids = list(
+            obj.filter(creator__isnull=False)
+            .values_list("creator", flat=True)
+            .distinct()
+            .order_by()
+        )
         insight.active_participants.add(*ids)
```
goapunk, line 88: @m4ra the issue was that we didn't exclude objects which have no creator (e.g. in the poll)

apps/projects/insights.py
```diff
@@ -68,16 +68,41 @@ def create_insight(project: Project) -> ProjectInsight:
...
+        # content from unregistered users doesn't have a creator but a content_id
+        if model_field_exists(obj.model, "content_id"):
+            content_ids = set(
```
goapunk, line 91: I now added more generic way of counting the unregistered participants, what do you think ? If that's fine I'd squash this commit and merge it afterwards

m4ra:
@goapunk LGTM. Go ahead with commits squashing and merge

PR #2833, opened 2024-11-05
apps/projects: check if a creator is set for a poll answer/vote to ac…

goapunk:
…commodate the new unregistered voting

**Tasks**
- [x] Changelog

depends on https://github.com/liqd/adhocracy4/pull/1683


m4ra (approved):
thanks!

PR #2847, opened 2024-12-03
[#8499] a4-polls: create info box for unregistered users

hom3mad3:
### Polls UI customization: Info box for unregistered users
- created new snippet for the box
- refactored the module description to remove duplicates and make it more maintainable

#### Desktop
Screenshot 2024-12-03 at 16-06-11 Private Project — adhocracy _liqd-orga

### DIFF
--- a/apps/projects/templates/a4_candy_projects/project_detail.html
+++ b/apps/projects/templates/a4_candy_projects/project_detail.html
@@ -196,16 +196,7 @@
                     </div>
 
                     {% else %} <!-- if just one module and no phase view to dispatch to -->
-                    <div class="row">
-                        <div class="col-md-8 offset-md-2 mb-3">
-                            {% block module_description %}
-                                {% if module.phases.first.type != 'a4_candy_interactive_events:issue' %}
-                                <h1 class="project-header__title">{{ module.name }}</h1>
-                                <p class="project-header__description">{{ module.description }}</p>
-                                {% endif %}
-                            {% endblock %}
-                        </div>
-                    </div>
+                    {% include "a4modules/includes/module_description.html" %}
                     {% if module.phases.first.type != 'a4_candy_interactive_events:issue' %}
                         {% if not module.active_phase %}
                         <div class="row">
--- a/adhocracy-plus/assets/scss/components/_info_box.scss
+++ b/adhocracy-plus/assets/scss/components/_info_box.scss
@@ -1,0 +1,21 @@
+.info-box {
+    background-color: $brand-primary-tint;
+    padding: $spacer;
+    margin: $spacer 0;
+    display: flex;
+    flex-direction: column;
+
+    &__content {
+        display: flex;
+        align-items: first baseline;
+        gap: $spacer; // Space between icon and text
+
+        p {
+            margin-bottom: 0;
+        }
+    }
+
+    &__text {
+        flex-grow: 1;
+    }
+}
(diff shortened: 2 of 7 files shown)


apps/projects/templates/a4_candy_projects/project_detail.html
```diff
@@ -1,324 +1,285 @@
...
-<meta property="og:image:height" content="500">
-{% endif %}
-{% endblock %}
+    {% if project.image %}
+        <meta name="twitter:image" content="{{ project.image |thumbnail_url:'heroimage' }}">
+            <meta name="linkedin:image" content="{{ project.image |thumbnail_url:'heroimage' }}">
+                <meta property="og:image" content="{{ project.image |thumbnail_url:'heroimage' }}">
+                    <meta property="og:image:width" content="1500">
+                        <meta property="og:image:height" content="500">
+                        {% endif %}
```
goapunk, line 13: the indentation here seems off, should be on the same line as the `if`, same goes for the blocks below

goapunk:
@hom3mad3 the design changed for me now, also compared to the screenshot you posted so I assume it's not intentional, the module details are now centered in a smaller container:
Screenshot 2024-12-05 at 16-48-00 poll — adhocracy _liqd-orga

goapunk:
> > @hom3mad3 the design changed for me now, also compared to the screenshot you posted so I assume it's not intentional (or a firefox vs safari issue?) The module details are now centred in a smaller container:
> 
> argh :/ very confused now

looks like the content is suddenly full width instead of having the grey column left and right, maybe something got lost during the reformatting?

apps/projects/templates/a4_candy_projects/project_detail.html
```diff
@@ -1,324 +1,296 @@
...
-                    <div class="row">
-                    {% if project.is_private %}
-                        <div class="col-md-10 col-lg-8 offset-md-1 offset-lg-2 u-font-size-xs my-4">
-                            <i class="fas fa-lock pe-2" aria-hidden="true"></i>{% translate 'This project is not publicly visible. Only invited users can see content and actively participate.' %}
-                        </div>
-                    {% elif project.is_semipublic %}
-                        <div class="col-md-10 col-lg-8 offset-md-1 offset-lg-2 u-font-size-xs my-4">
-                            <i class="fas fa-eye pe-2" aria-hidden="true"></i>{% translate 'This project is publicly visible. Invited users can actively participate.' %}
+                    <!-- these blocks are only filled when in phase view -->
+                    {% if project_action %}
```
goapunk, line 244: This won't work as far as I know, there is no good way to check if a `block` has content or not. It sadly just means that we will have to have the empty div below

goapunk (changes requested):
The poll header bit is fine now but sadly the grey side bits are gone:
aplus-dev (https://aplus-dev.liqd.net/buro-fur-vielfalt/projects/umfrage-3/):
Screenshot 2024-12-10 at 17-36-57 Poll — adhocracy _Julian-Test2244
and in this pr:
Screenshot 2024-12-10 at 17-37-17 poll — adhocracy _liqd-orga

goapunk:
> > The poll header bit is fine now but sadly the grey side bits are gone
> 
> @goapunk thanks. i'll probably just start this from scratch. all the context switching those last weeks means i have no idea what happened anymore 🫠

No worries, maybe a helpful approach especially for reviewing would be to either first do the changes without the formatting (e.g. commit with -n) or the other way round, first format the files and then do the changes in a second commit

goapunk (approved):
LGTM, thanks!

PR #2848, opened 2024-12-05
[#8443] a4-polls: adapt ui for textarea changes

hom3mad3:
UI fixes to adapt to https://github.com/liqd/adhocracy4/pull/1687


goapunk (approved):
LGTM, thanks. Could you add a commit which updates to the latest a4 so we have the changes from your a4 PR?

PR #2888, opened 2025-01-02
settings/base.py: enable new open poll feature by default

goapunk:
Enable the option to create open polls by default (I don't really see a reason not to allow it by default).

**Tasks**
- [ ] Changelog


goapunk:
I'll merge this so testing of the open poll can be resumed, if there's a good reason to not enable it by default we can have a follow up PR which sets it to False, it's good to document/have in the settings anyway

PR #2916, opened 2025-01-14
remove stylelint workaround

goapunk:
They already released a fixed version, so we don't need the workaround any more

**Tasks**
- [ ] Changelog


m4ra (approved):
Thanks!

PR #2933, opened 2025-02-13
[updates] specify python version

partizipation:
If a developer is using pyenv, a new `.python-version` file will force them onto the latest supported version. 

If not, they can now see in the README the supported version. Additionally, if they are on a non-supported version, when running `make install` an warning will be printed to the console.

@m4ra you mentioned `make clean` doesn't remove the venv directory when using pyenv, but it seems to be. Maybe this is no longer happening. If there are other issues with pyenv we could smooth out, maybe now would be a good time though, so let me know.


hom3mad3:
@arc64 this about the python version, btw

hom3mad3 (approved):
Nice!

.python-version
```diff
@@ -0,0 +1 @@
+3.11.6
```
m4ra, line 1: Personally I use pyenv for versions, and pyenv for virtualenv, so this file points to the name of my pyenv's virtualenv, e.g named `root` which is located in `.pyenv/versions/(venv)` and that's why cannot be deleted with the `make clean` command. I don't think this is inclusive enough approach.

m4ra (changes requested):
I would argue for writing a [project.toml](https://parottasalna.com/2025/02/12/learning-notes-71-pyproject-toml/) in the adhocracy-4 library and define the python version there, instead of constraining how developers should setup their virtualenv. Having a project.toml is also the standard fashion of publishing packages.

hom3mad3:
with a docker build this should also be fixed tbh 🙈

Issue #2171, opened 2022-12-21
Consistent tab behaviours


phillimorland:
**URL:** https://aplus-dev.liqd.net/teststadt/dashboard/organisations/projects/ https://aplus-dev.liqd.net/teststadt/projects/lorem/
**expected behaviour:** tabs nav have consistent behaviour 
**behaviour:** tabs in project can be navigated via arrow keys (best practice) but tabs in dashboard are navigated via tab key

PR #2937, opened 2025-03-27
Improve accessibility of dashboard tabs (issue #2171)

artyfakt:
I'm currently familiarizing myself with a4/a+ and thought I would try working on some open issues in the process. To start off, I had a stab at #2171.

- since dashboard tabs load new pages (instead of revealing some hidden content on the same page), I found that applying Bootstrap's built-in tab functionality doesn't work
- instead I added some javascript that can be loaded via the `extra_js` block in the django template
- the script ensures keyboard navigability in line with [ARIA recommendations for tabs with manual activation](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/) (Press tab to enter the tablist, press directional keys to navigate the tablist, press enter to activate a tab)
- for the script to work, the tablist needs to have the class "dashboard-tabs" and the `role="tablist"`
- also each tab of the tablist needs to have the `role="tab"`

I'm happy to work more on this according to your feedback. Also, let me know if there are any other good first issues (preferably frontend-heavy).

**Tasks**
- [ ] Changelog

### DIFF
--- a/adhocracy-plus/templates/a4dashboard/base_dashboard.html
+++ b/adhocracy-plus/templates/a4dashboard/base_dashboard.html
@@ -52,19 +52,31 @@
             {% url 'a4dashboard:newsletter-create' organisation_slug=view.organisation.slug as newsletter_create %}
             {% url 'a4dashboard:organisation-settings' organisation_slug=view.organisation.slug as organisation_settings %}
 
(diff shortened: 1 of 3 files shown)


t-j-q:
Hello @artyfakt thank you for contributing! I'm sorry that we hadn't have time to get back to you- we are quite occupied currently. Maybe @shn-liqd or @partizipation could have a look? 

Best wishes from Berlin
Tietje

PR #3066, opened 2026-04-13
apps/landingpage: Make USP blocks clickable

shn-liqd:


glanzel (changes requested):
Could be merged after the linter errors are solved

PR #3174, opened 2026-08-05
[ST-2529] Notifications bugs

partizipation:
Three bugs fixed:
1. Notifications task was running every 72 hours but only looking 24 hours ahead/back, so some phase start/end and upcoming event notifications were getting missed
2. User setting for newsletter opt-in was getting cleared out by missing field on dashboard form
3. After guest user convert to regular user, notifications were not getting sent because user was still being identified as guest due to guest user row not being deleted 

**Tasks**
- [x] Changelog

:.::..:::::: [Aligned Asana Task](https://app.asana.com/1/1175467295883992/project/1208559589495764/task/1216906205492170) ::.:.::.::::


apps/notifications/services.py
glanzel, line 151: are Notification Object are created for Email only Notifications aswell ?
Or is that not wanted ?

glanzel (approved):
Seem okay. I have one Question which does not regard the fixes but the overall logic.
Could be merged when test is fixed.

apps/notifications/services.py
partizipation: notification rows are supposed to only be created for in-app, not for email

PR #3187, opened 2026-08-19
deps: use typescript

shn-liqd:


apps/organisations/assets/react_language_choice.tsx
```diff
@@ -142,11 +133,6 @@ const LanguageChoice = (props) => {
                     <span key={languagecode}>
                       {activeLanguages.indexOf(languagecode) !== -1 &&
                         <button
-                          href={
```
shn-liqd, line 145: todo: check

shn-liqd: ✅ Language toggle working fine

glanzel (approved):
I checked it out locally and tested if it installs and runs properly.
I did not read all files.

PR #3188, opened 2026-08-19
e2e tests

shn-liqd:


.github/workflows/django.yml
```diff
@@ -124,3 +124,65 @@ jobs:
...
+          uv pip install -r requirements/dev.txt --system
+      - name: Install Frontend Dependencies
+        run: |
+          pnpm install --frozen-lockfile
+      - name: Build Frontend
+        run: |
+          pnpm run build
+      - name: Install Playwright Chromium
+        run: |
+          python -m playwright install --with-deps chromium
```
glanzel: only 21s but  could be optimized in terms of ci speed if a custom image is used.
not  relevant as long as it is faster than the main job.

.github/workflows/django.yml
```diff
@@ -124,3 +124,65 @@ jobs:
...
+      - name: Run End-to-End Tests
+        env:
+          DJANGO_ALLOW_ASYNC_UNSAFE: "1"
+        run: |
+          python manage.py collectstatic > /dev/null
+          py.test -m e2e --ds=adhocracy-plus.config.settings.e2e \
```
glanzel: maybe add  `-n auto` here?  but would probably need another db.

glanzel (approved):
just my thought.
No Change Request

.github/workflows/django.yml
shn-liqd: Thanks , implemented

